PR #10, opened 2015-01-22
Added support for login with auth.tc.net

jonasfj:
Added `taskcluster login` command that will open a browser to `auth.taskcluster.net` which after login and "OK" will redirect to localhost where the `taskcluster login` command picks up the temporary credentials and stores them in a file.


lightsofapollo:
r+ if you apply this patch which fixes the global form `taskcluster login` https://gist.github.com/lightsofapollo/32818064e6027ba5c63a

jonasfj:
@lightsofapollo, feel free to publish to npm... I don't think I have permission...

lightsofapollo:
@jonasfj you can now push to this project on npm ++

PR #15, opened 2015-08-10
Added interactive ssh-like support

jonasfj:
This should add `taskcluster ssh <taskId>` and act somewhat like ssh.

It resize with the terminal etc... enjoy...

### DIFF
--- a/bin/taskcluster-ssh.js
+++ b/bin/taskcluster-ssh.js
@@ -14,6 +14,7 @@ var yargs = require('yargs')
   .example('taskcluster ssh', 'taskId')
 
 var taskId = yargs.argv._[0];
+assert(taskId, "Missing taskId");
 
 var queue = new taskcluster.Queue();
 
--- a/bin/taskcluster-ssh.js
+++ b/bin/taskcluster-ssh.js
@@ -69,6 +69,6 @@ queue.status(taskId).then(function(result) {
     process.exit(code);
   });
 }).catch(function(err) {
-  console.log("Error: " + err.stack);
+  console.log(err.stack);
   process.exit(1);
 });
--- a/bin/taskcluster-ssh.js
+++ b/bin/taskcluster-ssh.js
@@ -1,0 +1,73 @@
+var taskcluster = require('taskcluster-client');
+var Promise = require('promise');
+var DockerClient = require('docker-exec-websocket-server').DockerExecClient;
+var https = require('https');
+var url = require('url');
+var _ = require('lodash');
+var assert = require('assert');
+require('../lib/config').load();
+
+var INTERACTIVE_SOCKET_NAME = 'private/docker-worker/interactive.sock';
+
+var yargs = require('yargs')
+  .usage('SSH into task container')
+  .example('taskcluster ssh', 'taskId')
+
+var taskId = yargs.argv._[0];
+
+var queue = new taskcluster.Queue();
+
+queue.status(taskId).then(function(result) {
+  // Get run and check state
+  var run = _.last(result.status.runs);
+  console.log("Latest run: " + run.runId);
+  if (run.state !== 'running') {
+    console.log("Task must be running for us to connect!");
+    process.exit(1);
+  }
+
+  // Build URL for the interactive socket
+  var signedUrl = queue.buildSignedUrl(
+    queue.getArtifact, taskId, run.runId, INTERACTIVE_SOCKET_NAME
+  );
+
+  var req = https.get(url.parse(signedUrl));
+  return new Promise(function(accept, reject) {
+    req.on('response', accept);
+    req.on('error', reject);
+  });
+}).then(function(res) {
+  assert(res.statusCode === 303, "Expected a 303 redirect");
+  var socketUrl = res.headers.location;
+
+  var client = new DockerClient({
+    url: socketUrl,
+    tty: true,
+    command: ['/.taskclusterutils/busybox', 'sh'],
+  });
+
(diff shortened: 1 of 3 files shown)


gregarndt:
Awesome, I've put this on my radar.  When I test out some interactive stuff later today, I'll check this out.

bin/taskcluster-ssh.js
```diff
@@ -0,0 +1,73 @@
...
+var assert = require('assert');
+require('../lib/config').load();
+
+var INTERACTIVE_SOCKET_NAME = 'private/docker-worker/interactive.sock';
+
+var yargs = require('yargs')
+  .usage('SSH into task container')
+  .example('taskcluster ssh', 'taskId')
+
+var taskId = yargs.argv._[0];
```
gregarndt: If a task ID is not provided, this results in a taskcluster-client error.  Could task id be required and some informative text outputted to the user?

bin/taskcluster-ssh.js
```diff
@@ -0,0 +1,73 @@
...
+    return new Promise(function(accept, reject) {
+      client.on('exit', accept);
+      client.on('error', reject);
+    });
+  }).then(function(code) {
+    client.close();
+    process.exit(code);
+  });
+}).catch(function(err) {
+  console.log("Error: " + err.stack);
```
gregarndt: Prefixing a thrown error displays as "Error: Error:".  Perhaps we don't need it.

bin/taskcluster-ssh.js
```diff
@@ -0,0 +1,73 @@
...
+
+    return new Promise(function(accept, reject) {
+      client.on('exit', accept);
+      client.on('error', reject);
+    });
+  }).then(function(code) {
+    client.close();
+    process.exit(code);
+  });
+}).catch(function(err) {
```
gregarndt: any value in trying to determine some of the errors ahead of time?  If an invalid task ID is given (such as missing a character or pasting in a graph ID instead of task ID), the thrown error is "Error: Malformed URL parameters" which doesn't help much in the way of fixing the issue.

bin/taskcluster-ssh.js
```diff
@@ -0,0 +1,73 @@
...
+var taskId = yargs.argv._[0];
+
+var queue = new taskcluster.Queue();
+
+queue.status(taskId).then(function(result) {
+  // Get run and check state
+  var run = _.last(result.status.runs);
+  console.log("Latest run: " + run.runId);
+  if (run.state !== 'running') {
+    console.log("Task must be running for us to connect!");
```
gregarndt: I wonder if we could have an option at some point to wait for this artifact to be present so that way we can leave the cli trying to connect and not miss the timeout for an interactive session

gregarndt:
I'm not sure the cause yet, but I was using this with a running container doing an emulator build and it became extremely painful and very slow to respond.  Not sure if ti's the machine, docker websocket, or something wrong on my end.

bin/taskcluster-ssh.js
```diff
@@ -0,0 +1,73 @@
...
+  var client = new DockerClient({
+    url: socketUrl,
+    tty: true,
+    command: ['/.taskclusterutils/busybox', 'sh'],
+  });
+
+  return client.execute().then(function() {
+    // Set terminal size and update when terminal is resized
+    client.resize(process.stdout.rows, process.stdout.columns);
+    process.stdout.on('resize', function() {
```
gregarndt: I'm not sure if I'm doing something wrong on my end, but when I resized my terminal, it doesn't look like it adjusted it on the receiving side.  Top pane here is the docker websocket session, and below that is an ssh session.  Notice that $MANIFEST is cut off in the top pane.
screen shot 2015-08-10 at 10 50 58 am

bin/taskcluster-ssh.js
jonasfj: Nit, we can improve error handling some other day...

bin/taskcluster-ssh.js
jonasfj: We could do that...
Let's leave it for a later PR.

bin/taskcluster-ssh.js
jonasfj: it seems to work correctly on my end... you might have to debug this locally...
Make sure you try to use `bash` instead of just `sh` as is default...

It's also possible that this is a vim specific bug... I didn't try vim.

bin/taskcluster-ssh.js
jonasfj: Added  an assert... no ideal, but better than nothing...

jonasfj:
@gregarndt,
I'm not going to invest a lot in cleaning this up, as cli tools already seem pretty messy.
If you want to add better error handling etc, feel free... but afaik we should refactor to babel, etc. before we do any of that...

I mostly did this PR, so that my work on terminal integration wouldn't go lost...

bin/taskcluster-ssh.js
gregarndt: This is not specific to vim.  See word wrap below.  
screen shot 2015-08-10 at 1 08 13 pm

jonasfj: @gregarndt is the resize feature rolled out serverside?
Because yeah, it seems to not really work...

It does when I use a local server... that I hacked up... hmm... I think..

gregarndt: as @jonasfj confirmed on IRC, this is not on the worker type that was being used.  This is a nonissue.

PR #17, opened 2016-08-02
[WIP] Go tc cli

jonasfj:
@jhford, @petemoore, I discussed taskcluster-cli with you and promised to start a draft... Here we go...

This has absolutely zero tests and I haven't even tried to use it... It's just sort of a start... 
Feel free to help out...

My hope is to hook this up to `taskcluster-github` and have it rebuilding on push and latest task index in the index... then use https://github.com/inconshreveable/go-update so we can have automatic updates...
With the signing key store in `taskcluster-secrets`...

Note: I suspect we would need a build without auto-updater for embedding in various kind of images...


config/config.go
```diff
@@ -0,0 +1,350 @@
...
+		if option == nil {
+			value = config
+		}
+
+		// Write output
+		if _, err := out.Write(formatter(value)); err != nil {
+			fmt.Println("Error writing result, error: ", err)
+			return false
+		}
+	}
```
petemoore: See [Effective Go](https://golang.org/doc/effective_go.html#switch). This should be

```
switch {
  case argv["help"]:
    ...
  case argv["set"]:
    ...
  case argv["reset"]:
    ...
  default:
    ...
}

File: config/config.go [hunk 2]
```diff
@@ -0,0 +1,350 @@
...
+	"errors"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"sort"
+	"strings"
+
+	"github.com/taskcluster/taskcluster-cli/extpoints"
+	"github.com/taskcluster/taskcluster-client-go/tcclient"
```
petemoore: This should be `tcclient "github.com/taskcluster/taskcluster-client-go"`. You probably need to run `go get -u` as it looks like you've developed against an older version of the client.

client/credentials.go
```diff
@@ -0,0 +1,108 @@
...
+	URL += "?bewit=" + url.QueryEscape(a.Bewit())
+	return URL, nil
+}
+
+// SignRequest will add an Authorization header
+func (c *Credentials) SignRequest(req *http.Request, hash hash.Hash) error {
+	s, err := c.SignHeader(req.Method, req.URL.String(), hash)
+	req.Header.Set("Authorization", s)
+	return err
+}
```
petemoore: Isn't all of this already implemented in `tcclient` package?

client/credentials.go
```diff
@@ -0,0 +1,108 @@
...
+	// Create a hawk auth
+	a, err := hawk.NewURLAuth(url, &hawk.Credentials{
+		ID:   c.ClientID,
+		Key:  c.AccessToken,
+		Hash: sha256.New,
+	}, 0)
+	if err != nil {
+		return nil, err
+	}
+	a.Method = method
```
petemoore: You also need to set a `nonce`...

```
    b := make([]byte, 8)
    _, err = io.ReadFull(rand.Reader, b)
    if err != nil {
        panic(err)
    }   
    a.Nonce = base64.StdEncoding.EncodeToString(b)[:8]
```

I even tested to be sure that `newAuth` isn't setting one, and confirmed it isn't:

```
package main

import (
        "fmt"
        "net/http"
        "net/http/httputil"
        "os"

        "github.com/taskcluster/taskcluster-cli/client"
)

func main() {
        c := &client.Credentials{
                ClientID:    os.Getenv("TASKCLUSTER_CLIENT_ID"),
                AccessToken: os.Getenv("TASKCLUSTER_ACCESS_TOKEN"),
                Certificate: os.Getenv("TASKCLUSTER_CERTIFICATE"),
        }
        req, _ := http.NewRequest("GET", "https://secrets.taskcluster.net/v1/secrets", nil)
        c.SignRequest(req, nil)
        req.Header.Write(os.Stdout)
        resp, _ := (&http.Client{}).Do(req)
        data, _ := httputil.DumpResponse(resp, true)
        fmt.Println(string(data))
}
```

gives me:

```
Authorization: Hawk id="mozilla-ldap/pmoore@mozilla.com/dev", mac="************", ts="1470333769", nonce=""
HTTP/1.1 200 OK
Content-Length: 19
Access-Control-Allow-Headers: X-Requested-With,Content-Type,Authorization,Accept,Origin
Access-Control-Allow-Methods: OPTIONS,GET,HEAD,POST,PUT,DELETE,TRACE,CONNECT
Access-Control-Allow-Origin: *
Access-Control-Request-Method: *
Connection: keep-alive
Content-Type: application/json; charset=utf-8
Date: Thu, 04 Aug 2016 18:02:50 GMT
Etag: W/"13-575462c4"
Server: Cowboy
Via: 1.1 vegur
X-Powered-By: Express

{
  "secrets": []
}
```

(see - no `nonce` and no secrets)

However, if I add my patch above, I get the correct output:

```
Authorization: Hawk id="mozilla-ldap/pmoore@mozilla.com/dev", mac="*******", ts="1470333645", nonce="2Bd7/MZ4"
HTTP/1.1 200 OK
Content-Length: 1468
Access-Control-Allow-Headers: X-Requested-With,Content-Type,Authorization,Accept,Origin
Access-Control-Allow-Methods: OPTIONS,GET,HEAD,POST,PUT,DELETE,TRACE,CONNECT
Access-Control-Allow-Origin: *
Access-Control-Request-Method: *
Connection: keep-alive
Content-Type: application/json; charset=utf-8
Date: Thu, 04 Aug 2016 18:00:47 GMT
Etag: W/"oV4NVLg3rFx8j5I+ZVD99g=="
Server: Cowboy
Via: 1.1 vegur
X-Powered-By: Express

{
  "secrets": [
    "garbage/pmoore/.......",
    "project/releng/......",
    "project/releng/......",
    ....
    "project/taskcluster/.......",
    "project/taskcluster/.......",
    ....
  ]
}
```

so it looks like server side we require the nonce, so ... you'll need to add it. ;)

petemoore: Probably would be good to have at least _one_ integration test, to confirm the signing works as expected...

petemoore: OK - my PR got merged, so this is now fixed upstream! 👯 

https://github.com/tent/hawk-go/pull/9

petemoore: @jonasfj OK I realised my upstream fix was actually incorrect, and it needs to be fixed in this library, so actually, you'll need to refactor a bit. You need a nonce in the case of signing a request, and you shouldn't have a nonce in the case of signing a request.

So you need to split this up and call `hawk.NewURLAuth` inside `SignURL` and `hawk.NewRequestAuth` inside `SignRequest` instead of exclusively using `hawk.NewURLAuth` in both.

petemoore: @jonasfj just a reminder that this is waiting on you! :-)

config/config.go
jonasfj: Obviously the answer here is to split into methods...  But yes, agree and don't care :)

config/config.go
jonasfj: clearly `godep restore` has been playing games with my clones...

client/credentials.go
jonasfj: yes, but `tcclient` isn't very flexible... Doesn't allow sign header for urls... And forces me to use the method names.

Note: this was just easier, or so I hope, particularly for the case where I want to auto-generate all cli commands for all api methods.

client/credentials.go
jonasfj: Nice catch.. I'll just set nonce... there is hardly any harm in setting it manyally...

jonasfj:
Forced pushed to master as this is the way we want to go... and good first bugs are now out...

Issue #19, opened 2016-09-15
Add version command


walac:
In summary, this will add `taskcluster version` command. It can be just a template with a fake command version until we decide how we store and update the version.

nikinath:
@walac Hello, I am an outreachy participant and would like to work on this bug. Can you please assign it to me?
Thank you.

PR #26, opened 2016-09-19
Added version command

nikinath:
This adds the `taskcluster version` command. Additionally, the `taskcluster --version` command can be used to display the version.

Currently, a fake version is printed. This can be changed later.

A new `version.go` file is added. The version is specified in `docopt.Parse()` as per the [documentation](https://godoc.org/github.com/docopt/docopt.go#Parse). 

This PR fixes issue(#19).

/cc @walac for review.

### DIFF
--- a/main.go
+++ b/main.go
@@ -12,6 +12,7 @@ import (
 	"github.com/taskcluster/taskcluster-cli/client"
 	"github.com/taskcluster/taskcluster-cli/config"
 	"github.com/taskcluster/taskcluster-cli/extpoints"
+	"github.com/taskcluster/taskcluster-cli/version"
 )
 
 func pad(s string, length int) string {
@@ -64,21 +65,14 @@ func main() {
 	usage += "\n"
 
 	// Parse arguments
-	arguments, _ := docopt.Parse(usage, nil, true, "TaskCluster version 1.0", true)
+	arguments, _ := docopt.Parse(usage, nil, true, version.VersionNumber, true)
 	cmd := arguments["<command>"].(string)
 	args := arguments["<args>"].([]string)
 
 	// Special case for handling "taskcluster help" ensuring it's the same as
-	// "taskcluster --help".
+	// "taskcluster --help". This should be the only special case necessary!
 	if cmd == "help" && len(args) == 0 {
 		fmt.Print(usage)
-		return
-	}
-
-	// Special case for handling "taskcluster version" ensuring it's the same as
-	// "taskcluster --version".
-	if cmd == "version" && len(args) == 0 {
-		fmt.Println("TaskCluster version 1.0")
 		return
 	}
 
@@ -94,7 +88,7 @@ func main() {
 	// Parse args for command provider
 	subArguments, _ := docopt.Parse(
 		provider.Usage(), append([]string{cmd}, args...),
-		true, "TaskCluster version 1.0", false,
+		true, version.VersionNumber, false,
 	)
 
 	// Create credentials, if available in configuration
--- a/main.go
+++ b/main.go
@@ -64,14 +64,21 @@ func main() {
 	usage += "\n"
 
 	// Parse arguments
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -64,17 +64,24 @@ func main() {
 	usage += "\n"
 
 	// Parse arguments
-	arguments, _ := docopt.Parse(usage, nil, true, "taskcluster", true)
+	arguments, _ := docopt.Parse(usage, nil, true, "TaskCluster version 1.0", true)
```
walac: `taskcluster version` and `taskcluster --version` should read the version number from the same location, even for a fake version.

main.go
```diff
@@ -64,17 +64,24 @@ func main() {
...
 		return
 	}
 
+	// Special case for handling "taskcluster version" ensuring it's the same as
+	// "taskcluster --version".
+	if cmd == "version" && len(args) == 0 {
+		fmt.Println("TaskCluster version 1.0")
+		return
+	}
+
```
walac: `extpoints` handles subcommands, this piece of code is unnecessary.

main.go
```diff
@@ -87,7 +94,7 @@ func main() {
 	// Parse args for command provider
 	subArguments, _ := docopt.Parse(
 		provider.Usage(), append([]string{cmd}, args...),
-		true, "taskcluster", false,
+		true, "TaskCluster version 1.0", false,
```
walac: Please, provide version number from a single location.

version.go
```diff
@@ -0,0 +1,40 @@
...
+func (version) Summary() string {
+	return "Prints the TaskCluster version."
+}
+
+func (version) Usage() string {
+	usage := "Prints the TaskCluster version\n"
+	usage += "\n"
+	usage += "Usage:\n"
+	usage += "  taskcluster version\n"
+	return usage
```
walac: nit: use \`\` strings, like [here](https://github.com/taskcluster/taskcluster-worker/blob/master/commands/daemon/daemon.go#L29)

version.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+func (version) Execute(context extpoints.Context) bool {
+	command := context.Arguments["<command>"].(string)
+	provider := extpoints.CommandProviders()[command]
+	if provider == nil {
+		fmt.Println("Unknown command: ", command)
```
walac: nit: as this should never happen, replace this `Println` by [`panic`](https://gobyexample.com/panic)

walac:
One thing I forgot to ask, move `version.go` to its own `version` directory :)

main.go
jonasfj: hmm... If we write in capital it's the name and the name of the project is "TaskCluster CLI" or something like that.
the lowercase "taskcluster" was intended to be the name of the binary.

Hmm... are there any GNU guidelines for version string formatting (I can't see any consistency between, less, vim, curl, wget, man, gpg, so I'm guessing no. Maybe do a search through GNU manuals.. Or jsut go with something like `taskcluster (TaskCluster CLI) version %d.%d.%d` with `fmt.Sprintf`

Yes, designing a version message is totally bike-shredding :) Just nice to have fun stuff..

main.go
jonasfj: Yeah, it should be possible to add a subcommand that implement this command.

version/version.go
```diff
@@ -0,0 +1,44 @@
...
+	return nil
+}
+
+func (version) Summary() string {
+	return "Prints the TaskCluster version."
+}
+
+func usage() string {
+	return `Usage:
+  taskcluster VERSION
```
nikinath: According to the [docopt documentation](https://github.com/docopt/docopt#usage-pattern-format), the arguments are specified as upper case words. This forced me to make the usage as `taskcluster VERSION` instead of `taskcluster version`. 

Is that all right or am I missing something here?

walac: That's fine. You could also use `<version>`, but this is ok either.

Issue #21, opened 2016-09-15
Add test case for the credentials module


walac:
Taskcluster uses [Hawk]() for authentication under the hood. The [credentials module]() provides an interface for taskcluster authentication. Unfortunately, it doesn't have any unit test case. Implementing a test case for follows these steps:

* Create a `Credentials` object with `ClientID = tester` and `AccessToken = no-secret`
* Create a request to [testAuthenticate](https://docs.taskcluster.net/reference/platform/auth/api-docs#testAuthenticate)
* Sign it using `SignRequest`
* Perform the request and check it was successful

nanjekyejoannah:
Hello ...

Kindly assign me this issue..thanks

walac:
@nanjekyejoannah I sent you an invitation to taskcluster-cli-contributors team, once you accept the invite, I assign you to the bug

PR #27, opened 2016-09-20
Add test case for the credentials module

nanjekyejoannah:
I have added a test case to test the [credentials module](https://github.com/taskcluster/taskcluster-cli/issues/21)

### DIFF
--- a/client/credentials_test.go
+++ b/client/credentials_test.go
@@ -1,0 +1,69 @@
+package client
+
+import (
+	"testing"
+	"strconv"	
+)
+
+func TestCredentials(t *testing.T){
+
+	cred := &Credentials{
+		ClientID          : "tester"
+		AccessToken       : "no-secret"
+	}
+
+	var credentialsTests = []struct {
+		meth string
+		url  string
+		host string
+		port int
+		hdr  string
+		now  int64
+		perr error
+		verr error
+		key  string
+		hash func() hash.Hash
+		rply bool
+	}
+	{
+		{
+			hdr:  `Hawk id="1", ts="1353788437", nonce="k3j4h2", mac="zy79QQ5/EYFmQqutVnYb73gAc/U=", ext="hello"`,
+			hash: sha1.New,
+		},
+	}
+
+	for k, test := range credentialsTests {
+			if test.meth == "" {
+				test.meth = "GET"
+			}
+			if test.url == "" {
+				test.url = "/resource/4?filter=a"
+			}
+			if test.host == "" {
+				test.host = "example.com"
+			}
+			if test.port == 0 {
+				test.port = 8080
+			}
+			if test.now == 0 {
+				test.now = 1353788437
+			}
+			if test.hash == nil {
+				test.hash = sha256.New
+			}
+			if test.key == "" {
+				test.key = "werxhqb98rpaxn39848xrunpaw3489ruxnpa98w4rxn"
+			}
+	}
+
+	request := &http.Request{
+				Method:     test.meth,
+				RequestURI: test.url,
+				Host:       test.host + ":" + strconv.Itoa(test.port),
+				Header:     http.Header{"Authorization": {test.hdr}},
+			}
+
+	cred.SignRequest(request,test.hash)
+	auth, errors := cred.newAuth(test.meth , test.url,test.hash)
+	t.assert(errors,nil)
+}
--- a/client/credentials_test.go
+++ b/client/credentials_test.go
@@ -65,5 +65,5 @@ func TestCredentials(t *testing.T){
 
 	cred.SignRequest(request,test.hash)
 	auth, errors := cred.newAuth(test.meth , test.url,test.hash)
-	t.assert(errors,nil)
+	t.assert(errors,Equals,nil)
 }
--- a/client/credentials_test.go
+++ b/client/credentials_test.go
@@ -2,18 +2,21 @@ package client
 
 import (
 	"testing"
-	"strconv"	
+	"strconv"
+	"github.com/stretchr/testify/assert"	
 )
 
 func TestCredentials(t *testing.T){
 
-	cred := &Credentials{
-		ClientID          : "tester"
-		AccessToken       : "no-secret"
+	 assert := assert.New(t)
+
+	credentials := &Credentials{
+		ClientID          : "tester" ,
+		AccessToken       : "no-secret",
 	}
 
 	var credentialsTests = []struct {
-		meth string
+		method string
 		url  string
 		host string
 		port int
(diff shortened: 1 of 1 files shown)


client/credentials_test.go
```diff
@@ -0,0 +1,69 @@
...
+			}
+			if test.now == 0 {
+				test.now = 1353788437
+			}
+			if test.hash == nil {
+				test.hash = sha256.New
+			}
+			if test.key == "" {
+				test.key = "werxhqb98rpaxn39848xrunpaw3489ruxnpa98w4rxn"
+			}
```
walac: Why don't you initialize this on array creation?

client/credentials_test.go
```diff
@@ -0,0 +1,69 @@
...
+				Method:     test.meth,
+				RequestURI: test.url,
+				Host:       test.host + ":" + strconv.Itoa(test.port),
+				Header:     http.Header{"Authorization": {test.hdr}},
+			}
+
+	cred.SignRequest(request,test.hash)
+	auth, errors := cred.newAuth(test.meth , test.url,test.hash)
+	t.assert(errors,Equals,nil)
+}
```
walac: Please, format the source file with [goimports](https://godoc.org/golang.org/x/tools/cmd/goimports)

client/credentials_test.go
nanjekyejoannah: I think that makes much sense.

walac: This:
```go
var credentialsTests = []struct {
      meth string
      url  string
      host string
      port int
      hdr  string
      now  int64
      perr error
      verr error
      key  string
      hash func() hash.Hash
      rply bool
  }
{
      {
          meth: "GET",
          url: "/resource/4?filter=a",
          host: "example.com",
          port: 8080,
          hdr:  `Hawk id="1", ts="1353788437", nonce="k3j4h2", mac="zy79QQ5/EYFmQqutVnYb73gAc/U=", ext="hello"`,
          now: 1353788437,
          hash: sha1.New,
          key: "werxhqb98rpaxn39848xrunpaw3489ruxnpa98w4rxn"
      },
  }

  for k, test := range credentialsTests {
      // ...
  }
```

* Is more compact
* Decreases the [cyclomatic complexity](https://en.wikipedia.org/wiki/Cyclomatic_complexity) of the code
* It is a [declarative syntax](https://en.wikipedia.org/wiki/Declarative_programming), which is easier to read.

@jonasfj thoughts here?

client/credentials_test.go
```diff
@@ -0,0 +1,69 @@
...
+		url  string
+		host string
+		port int
+		hdr  string
+		now  int64
+		perr error
+		verr error
+		key  string
+		hash func() hash.Hash
+		rply bool
```
jonasfj: I would try to avoid names like this.. Just use  `method`, `reply`, `header`, they aren't that long, I understand how `verr` makes sense..

jonasfj:
hmm... I'm not sure I follow exactly that this tries to do...
Take a look at: https://docs.taskcluster.net/reference/platform/auth/api-docs#testAuthenticate
The idea is that `POST https://auth.taskcluster.net/v1/test-authenticate` can be authenticated with clientId `tester` and accessToken `no-secret`... So we can use this to test that header signing works..
`

client/credentials_test.go
```diff
@@ -0,0 +1,69 @@
...
+	request := &http.Request{
+				Method:     test.meth,
+				RequestURI: test.url,
+				Host:       test.host + ":" + strconv.Itoa(test.port),
+				Header:     http.Header{"Authorization": {test.hdr}},
+			}
+
+	cred.SignRequest(request,test.hash)
+	auth, errors := cred.newAuth(test.meth , test.url,test.hash)
+	t.assert(errors,Equals,nil)
```
jonasfj: I'm missing something here... (I'll admit I haven't been using `t.assert` much), but where does `Equals` come from?

nanjekyejoannah:
Reviews noted..will make a commit resolving tomorrow.

client/credentials_test.go
```diff
@@ -0,0 +1,54 @@
+package client
+
+import (
+	"testing"
+	"strconv"
+	"github.com/stretchr/testify/assert"
```
walac: This line has some trailing spaces...

client/credentials_test.go
```diff
@@ -0,0 +1,54 @@
...
+				Method:     credentialsTests.method,
+				RequestURI: credentialsTests.url,
+				Host:       credentialsTests.host + ":" + strconv.Itoa(credentialsTests.port),
+				Header:     http.Header{"Authorization": {credentialsTests.hdr}},
+			}
+
+	credentials.SignRequest(request,credentialsTests.hash)
+	auth, errors := credentials.newAuth(credentialsTests.method ,credentialsTests.url,credentialsTests.hash)
+	assert.Nil(errors)
+}
```
walac: hrm, I am using go 1.7. When I run `go test`, I get:

```shell
# github.com/taskcluster/taskcluster-cli/client
credentials_test.go:18:25: expected expression
credentials_test.go:33:10: expected 1 expression
credentials_test.go:34:8: illegal label declaration
credentials_test.go:54:2: expected '}', found 'EOF'
FAIL	github.com/taskcluster/taskcluster-cli/client [setup failed]
```

nanjekyejoannah:
I have ironed out the big issues but I think am still not getting a passing test any pointers.

client/credentials_test.go
```diff
@@ -0,0 +1,61 @@
...
+			now:    1353788437,
+			key:    "werxhqb98rpaxn39848xrunpaw3489ruxnpa98w4rxn",
+		},
+	}
+
+	request := &http.Request{
+		Method:     credentialsTests[0].method,
+		RequestURI: credentialsTests[0].url,
+		Host:       credentialsTests[0].host + ":" + strconv.Itoa(credentialsTests[0].port),
+		Header:     http.Header{"Authorization": {credentialsTests[0].hdr}},
```
walac: I think it is better to use [NewRequest](https://golang.org/pkg/net/http/#NewRequest) function. The way you are doing, we panic because `URL` property is not set.

client/credentials_test.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+	if credentialsTests[0].hash != nil {
+		panic(credentialsTests[0].hash)
+	}
```
walac: `hash` is expected to be non `nil`, isn't it?

nanjekyejoannah:
This should work now.

client/credentials_test.go
```diff
@@ -0,0 +1,56 @@
...
+func TestCredentials(t *testing.T) {
+
+	assert := assert.New(t)
+
+	credentials := &Credentials{
+		ClientID:    "tester",
+		AccessToken: "no-secret",
+	}
+
+	var credentialsTests = []struct {
```
walac: We have only one test case and index is hard coded, there is no point in have a array here.

client/credentials_test.go
```diff
@@ -0,0 +1,56 @@
...
+			port:   8080,
+			now:    1353788437,
+			key:    "werxhqb98rpaxn39848xrunpaw3489ruxnpa98w4rxn",
+		},
+	}
+
+	request, _ := http.NewRequest(credentialsTests[0].method, credentialsTests[0].url, nil)
+	credentials.SignRequest(request, credentialsTests[0].hash)
+	auth, errors := credentials.newAuth(credentialsTests[0].method, credentialsTests[0].url, credentialsTests[0].hash)
+	auth.Timestamp = time.Unix(1475317496, 0)
```
walac: This shouldn't use `credentials[0].now`?

client/credentials_test.go
```diff
@@ -0,0 +1,56 @@
...
+	auth.MAC = []byte("Jcelngt+a8loOSi7f7M9vCgdxBsXT4o+6kwkEqSMONg=")
```
walac: Use `credentialsTests[0].key`?

client/credentials_test.go
```diff
@@ -0,0 +1,56 @@
...
+	assert.Equal(nil, errors)
+	assert.Equal(auth.RequestHeader(), `Hawk id="tester", mac="mmUrSFCwMjlJ2rOwBhPoiVAhBuSvJX07gKwCPA8pdSE=", ts="1475317496", nonce="", hash="cYU8YZemp/Ii2w8ZeMfLIyuHxe4="`)
```
walac: `credentialsTests[0].hdr`?

client/credentials_test.go
```diff
@@ -0,0 +1,61 @@
...
+		perr   error
+		verr   error
+		key    string
+		hash   hash.Hash
+		reply  bool
+		nonce  string
+		ext    string
+	}
+
+	test_credentials := &CredentialsTests{
```
walac: Just a small nit: [golint](https://github.com/golang/lint) is complaining about using snake case variable name. Could you please rename it to `testCredentials`?

Issue #28, opened 2016-09-20
Add CLI command to upload/download to/from S3


jonasfj:
Using `auth.awsS3Credentials` takes taskcluster credentials and returns temporary STS credentials for S3.
It would be cool if we had a command like `taskcluster s3 copy ./myfile.txt s3://<bucket>/<prefix>`
that could get temp creds from auth and use those to upload a file to S3.
Obivously, it would have to do retries with exponential back-off, ideally also multi-part upload for large files.
Detect mimetype based on file extension... Provide content-md5 for S3 to verify.

See: https://docs.taskcluster.net/reference/platform/auth/api-docs#awsS3Credentials

Further steps could include:
- Download file
- Upload a folder of files
- Download a folder of files
- Adding extra headers like `x-amz-meta-content-sha256` for integrity checks on downloads

nikinath:
I'd like to work on this bug. Can I please get assigned?
Thank you!

djmitche:
I don't think we should implement this feature.  We currently have to manually whitelist all s3 buckets to which auth can grant access, which severely limits the utility of a feature like this.

PR #33, opened 2016-09-25
Added s3 command

nikinath:
The command can be used to upload/download file to AWS S3 using:
-  `taskcluster s3 upload <file> <region> <level> <bucket> <prefix>`
-  `taskcluster s3 download <key> <region> <level> <bucket> <prefix>`

I was (obviously) not able to test it since I don't have the required credentials (I get a HTTP 401 error). So what should be my next step to go about this?

Also, I faced one issue. `AWSS3CredentialsResponse` has an `Expires` field which returns  `tcclient.Time`. But `UploadInput` has an `Expires` field which takes `*time.Time` as mentioned [here](http://docs.aws.amazon.com/sdk-for-go/api/service/s3/s3manager/#UploadInput). I am a little confused about how to convert it. Any pointers on how to achieve this? 

The AWS SDK handles exponential backoff, multi part upload, etc.

This PR addresses issue(#28). /cc @walac

### DIFF
--- a/s3/s3.go
+++ b/s3/s3.go
@@ -2,6 +2,7 @@ package s3
 
 import (
 	"fmt"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 
@@ -9,7 +10,6 @@ import (
 	"github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/aws/aws-sdk-go/aws/session"
 	awsS3 "github.com/aws/aws-sdk-go/service/s3"
-	awsS3Manager "github.com/aws/aws-sdk-go/service/s3/s3manager"
 	"github.com/taskcluster/taskcluster-cli/config"
 	"github.com/taskcluster/taskcluster-cli/extpoints"
 	"github.com/taskcluster/taskcluster-client-go"
@@ -33,8 +33,8 @@ func (S3) Summary() string {
 func usage() string {
 	return `Upload/Download file to/from AWS S3.
 Usage:
-  taskcluster S3 upload <file> <region> <level> <bucket> <prefix>
-  taskcluster S3 download <key> <region> <level> <bucket> <prefix>
+  taskcluster S3 put <file> <bucket> <prefix>
+  taskcluster S3 get <key> <bucket> <prefix>
   taskcluster S3 help <subcommand>
 `
 }
@@ -55,23 +55,21 @@ func (S3) Execute(context extpoints.Context) bool {
 
 	// Print help for subcommands
 	if argv["help"] == true {
-
 		subcommand, ok := argv["<subcommand>"].(string)
 		if !ok {
 			panic(fmt.Sprintf("Unknown subcommand: %s", subcommand))
 		}
-
-		if subcommand == "upload" {
-			fmt.Println(`Usage:
-  taskcluster S3 upload <file> <region> <level> <bucket> <prefix>`)
-		} else if subcommand == "download" {
-			fmt.Println(`Usage:
-  taskcluster S3 download <key> <region> <level> <bucket> <prefix>`)
+		if subcommand == "put" {
+			fmt.Println("Usage:\ntaskcluster S3 put <file> <bucket> <prefix>")
+		} else if subcommand == "get" {
+			fmt.Println("Usage:\ntaskcluster S3 get <key> <bucket> <prefix>")
+		} else {
+			fmt.Printf("Invalid subcommand.\n")
 		}
 		return true
 	}
 
-	// Parse for bucket, prefix, level and region
+	// Parse for bucket and prefix
 	bucket, ok := argv["<bucket>"].(string)
 	if !ok {
 		fmt.Println("Invalid bucket format.")
@@ -82,15 +80,23 @@ func (S3) Execute(context extpoints.Context) bool {
 		fmt.Println("Invalid prefix format.")
 		return false
 	}
-	level, ok := argv["<level>"].(string)
-	if !ok {
-		fmt.Println("Invalid level format.")
-		return false
-	}
-	region, ok := argv["<region>"].(string)
-	if !ok {
-		fmt.Println("Invalid region format.")
-		return false
+
+	// Set level, filename and key
+	var level, filename, key string
+	if argv["put"] == true {
+		level = "read-write"
+		filename, ok = argv["<file>"].(string)
+		if !ok {
+			fmt.Println("Invalid file format.")
+			return false
+		}
+	} else if argv["get"] == true {
+		level = "read-write" // can change this later
+		key, ok = argv["<key>"].(string)
+		if !ok {
+			fmt.Println("Invalid key format.")
+			return false
+		}
 	}
 
 	// Load configuration
@@ -109,7 +115,7 @@ func (S3) Execute(context extpoints.Context) bool {
 		},
 	)
 
-	// Get credential for AWS S3
+	// Get credentials for AWS S3
 	resp, err := myAuth.AwsS3Credentials(level, bucket, prefix)
 	if err != nil {
 		fmt.Println("Failed to load AWS S3 credentials: ", err)
@@ -120,7 +126,6 @@ func (S3) Execute(context extpoints.Context) bool {
 	aws_access_key_id := resp.Credentials.AccessKeyID
 	aws_secret_access_key := resp.Credentials.SecretAccessKey
 	aws_session_token := resp.Credentials.SessionToken
-	//aws_expires := &resp.Expires
 
 	creds := credentials.NewStaticCredentials(aws_access_key_id, aws_secret_access_key, aws_session_token)
 	_, err = creds.Get()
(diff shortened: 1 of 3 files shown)


walac:
@nikinath Do you have a mozillians account? If so, you can use it to get taskcluster credentials. Reach us out at #tc-contributors if you need help.

jonasfj:
So the `<level>` in the API is `read-write` or `read-only` the level of access required from the STS creds returned by the API is implicit from the command `upload` or `download`.

Also, just nit-picking, maybe we should call the commands `get` and `put`.
I just love it when text aligns -- is there anything better, he he :)

s3/s3.go
```diff
@@ -0,0 +1,233 @@
...
+		fmt.Println("Failed to load configuration file, error: ", err)
+		return false
+	}
+
+	// Set credentials for auth
+	myAuth := auth.New(
+		&tcclient.Credentials{
+			ClientID:    config["config"]["clientId"].(string),
+			AccessToken: config["config"]["accessToken"].(string),
+			Certificate: config["config"]["certificate"].(string),
```
jonasfj: Won't this panic if there isn't a certificate? Or we do always set the empty certificate (I can't remember).

s3/s3.go
```diff
@@ -0,0 +1,233 @@
...
+	if !ok {
+		fmt.Println("Invalid prefix format.")
+		return false
+	}
+	level, ok := argv["<level>"].(string)
+	if !ok {
+		fmt.Println("Invalid level format.")
+		return false
+	}
+	region, ok := argv["<region>"].(string)
```
jonasfj: From docs: https://docs.taskcluster.net/reference/platform/auth/api-docs#awsS3Credentials
And the code: https://github.com/taskcluster/taskcluster-auth/blob/7be41bad7a37c405d36f63549dbee9224a2c823f/src/aws.js#L147

It's pretty clear that the policy we issue allows the action: `s3:GetBucketLocation`
How about we exploit that to automatically detect the region of the bucket :)

For a utility command it's really awesome to reduce number of unnecessary commands.
I imagine we could use this to upload things like a compiler to be used in automation.

s3/s3.go
```diff
@@ -0,0 +1,233 @@
...
+				return false
+			}
+		}
+
+		fmt.Printf("Successfully uploaded %s to s3://%s/%s\n", key, bucket, prefix)
+
+	} else if argv["download"] == true {
+
+		// Parse for key
+		key, ok := argv["<key>"].(string)
```
jonasfj: I suggest that the upload and download is split into separate functions... This just changes the code organization, but it could simplify the code a lot, reduce levels of indentation (improving code quality), and if we later decide to add sub-command like list, copy, delete then the code won't be horribly complicated.

 Maybe `Execute` should just parse arguments... Get credentials, detect bucket region and setup s3 client library... Then the file operations might be better to do in a separate method, maybe in a separate file.

jonasfj:
Generally, this looks really good! :)

s3/s3.go
nikinath: Oh yes, this simplifies a lot of things! For using `GetBucketLocation`, I need to set up a session (as can be seen [here](http://docs.aws.amazon.com/sdk-for-go/api/service/s3/#S3.GetBucketLocation)). This session in turn requires a region. I just specified a dummy region and it worked fine. I am a little confused as to why this works!

nikinath:
I uploaded and downloaded using this command and it works well! :)

I have a question about `level` - do you mean that if the command is `put/get`, the level is `read-write/read-only` respectively? I am unable to download if I specify the `level` that way since the scope I have requires the level to be `read-write`.

Maybe I'm missing something here?

jonasfj:
The method on auth which returns temporary STS creds for S3 gives the
credentials an IAM policy. Level=read-only and you can't do uploads,
level=read-write and you can both upload and download...
Hence, the download command implies level=read-only and the upload command
implies level=read-write.

Den 1. okt. 2016 1.47 PM skrev "Nikhita Raghunath" <notifications@github.com

nikinath:
Please correct me if I am wrong. 

If level = read-write or if level = read-only, downloads can take place. If I set level = read-only for the download command, I am unable to download since the scope I have is `auth:aws-s3:read-write:test-bucket-for-any-garbage/*` which requires the level to be `read-write`. Maybe the issue crops up because the level is involved in the scope?

s3/s3.go
```diff
@@ -0,0 +1,243 @@
...
+	var level, filename, key string
+	if argv["put"] == true {
+		level = "read-write"
+		filename, ok = argv["<file>"].(string)
+		if !ok {
+			fmt.Println("Invalid file format.")
+			return false
+		}
+	} else if argv["get"] == true {
+		level = "read-write" // can change this later
```
jonasfj: You should be able to fix this now...

s3/s3.go
```diff
@@ -0,0 +1,243 @@
...
+		key, ok = argv["<key>"].(string)
+		if !ok {
+			fmt.Println("Invalid key format.")
+			return false
+		}
+	}
+
+	// Load configuration
+	config, err := config.Load()
```
jonasfj: We shouldn't have to load config here.

Check out the `context` parameter...
https://github.com/taskcluster/taskcluster-cli/blob/091624c16d00e86d26cc53d2f5321bc010d56c7e/extpoints/interfaces.go#L20

s3/s3.go
```diff
@@ -0,0 +1,243 @@
...
+
+	creds := credentials.NewStaticCredentials(aws_access_key_id, aws_secret_access_key, aws_session_token)
+	_, err = creds.Get()
+	if err != nil {
+		fmt.Printf("Invalid credentials: %s", err)
+		return false
+	}
+
+	// Get bucket location
+	aws_config := aws.NewConfig().WithCredentials(creds).WithRegion("us-west-2")
```
jonasfj: If we have to hardcode a region for `GetBucketLocation` I suggest `us-east-1` that's the standard S3 region... I'm not sure if it makes any difference :)

s3/s3.go
```diff
@@ -0,0 +1,243 @@
...
+	svc := awsS3.New(sess)
+	params := &awsS3.GetBucketLocationInput{
+		Bucket: aws.String(bucket),
+	}
+	res, err := svc.GetBucketLocation(params)
+	if err != nil {
+		fmt.Println(err.Error())
+		return false
+	}
+	region := res.LocationConstraint
```
jonasfj: Perhaps we could put all this in a function, like `region := findBucketRegion(creds)`

jonasfj: well, `region, err := ...` probably...

s3/s3.go
```diff
@@ -0,0 +1,243 @@
...
+		return upload
+	} else if argv["get"] == true {
+		download := Get(svc, key, bucket, prefix)
+		return download
+	} else {
+		panic(fmt.Sprint("Invalid subcommand."))
+	}
+}
+
+func Put(svc *awsS3.S3, filename string, bucket string, prefix string) bool {
```
jonasfj: In go all capital identifiers are exported... This is just an internal function...

s3/s3.go
```diff
@@ -0,0 +1,243 @@
...
+
+	// Create a session which contains the configurations for the SDK.
+	// Use the session to create the service clients to make API calls to AWS.
+	aws_config = aws.NewConfig().WithCredentials(creds).WithRegion(*region)
+	sess = session.New(aws_config)
+	svc = awsS3.New(sess)
+
+	if argv["put"] == true {
+		upload := Put(svc, filename, bucket, prefix)
+		return upload
```
jonasfj: I think `return put(svc, filename, bucket, prefix)` would be acceptable.

s3/s3.go
```diff
@@ -0,0 +1,243 @@
...
+		fmt.Println("Invalid prefix format.")
+		return false
+	}
+
+	// Set level, filename and key
+	var level, filename, key string
+	if argv["put"] == true {
+		level = "read-write"
+		filename, ok = argv["<file>"].(string)
+		if !ok {
```
jonasfj: I don't think this can happen... docopt won't allow it.. if `argv["put"] == true` then the required arguments must also be there.

s3/s3.go
```diff
@@ -0,0 +1,243 @@
...
+			fmt.Println("Invalid file format.")
+			return false
+		}
+	} else if argv["get"] == true {
+		level = "read-write" // can change this later
+		key, ok = argv["<key>"].(string)
+		if !ok {
```
jonasfj: Same here, I don't think this can happen...

s3/s3.go
```diff
@@ -0,0 +1,243 @@
...
+
+	// Generate key from filename
+	key := filepath.Base(filename)
+
+	fmt.Printf("Uploading %s to s3://%s/%s...\n", key, bucket, prefix)
+
+	// Upload the file
+	params := &awsS3.PutObjectInput{
+		Bucket: aws.String(bucket),
+		Key:    aws.String(prefix + key),
```
jonasfj: hmm... I didn't think about the details here...

Maybe we should do it, such that we get:
```
$ taskcluster upload myfile.txt mybucket some/where             # myfile.txt ->  some/where
$ taskcluster upload myfile.txt mybucket some/where/myfile.txt  # myfile.txt -> some/where/myfile.txt
$ taskcluster upload myfile.txt mybucket some/where/            # myfile.txt -> some/where/myfile.txt
$ taskcluster upload myfile.txt mybucket some/where.txt         # myfile.txt -> some/where.txt
$ taskcluster upload myfile.txt mybucket some/where.txt/        # myfile.txt -> some/where.txt/myfile.txt
```

Ie. if the `prefix` ends in a slash then we add the `filepath.Base(filename)` otherwise, we don't add anything...
Similar to how `cp` works on linux... Sounds reasonable?

s3/s3.go
```diff
@@ -0,0 +1,243 @@
...
+	}
+
+	resp, err := svc.GetObject(params)
+	if err != nil {
+		fmt.Printf("Failed to download data to %s from s3://%s/%s\n", filename, bucket, prefix+key)
+		return false
+	}
+
+	// Write data to the file
+	data, err := ioutil.ReadAll(resp.Body)
```
jonasfj: This isn't smart if the body has a non-trivial size... a few GBs and this won't be fun...

`io.Copy(file, resp.Body)`

Also won't we need a retry loop here?

s3/s3.go
```diff
@@ -0,0 +1,243 @@
...
+
+func (S3) Summary() string {
+	return "Uploads/downloads file to/from AWS S3."
+}
+
+func usage() string {
+	return `Upload/Download file to/from AWS S3.
+Usage:
+  taskcluster S3 put <file> <bucket> <prefix>
+  taskcluster S3 get <key> <bucket> <prefix>
```
jonasfj: What is the difference between `<key>` and `<prefix>` both are addresses in s3, right?

I would think... `taskcluster S3 get <bucket> <prefix> [target]` might be better...
If `[target]` isn't given then `filepath.Join(os.Getwd(), path.Basename(prefix))` is used as filename...

s3/s3.go
jonasfj: Also we shouldn't do transformations of the prefix in here... We should do such transformation before we create the temporary STS creds... As we should ask for as little permissions as possible.

s3/s3.go
jonasfj: Yeah, and `nil` isn't a string so that'll panic.

s3/s3.go
jonasfj: My point is that there is no reason not to do `filename = args["<file>"].(string)` because if `argv["put"] == true` then the required arguments must be present, see:
https://github.com/taskcluster/taskcluster-cli/blob/master/config/config.go#L183
There we don't do the conditional check for `filename, ok = argv["<file>"].(string)`, because the `ok` check is unnecessary.

s3/s3.go
```diff
@@ -0,0 +1,250 @@
...
+			fmt.Println("Usage:\ntaskcluster S3 get <bucket> <prefix> [<target>]")
+		} else {
+			fmt.Printf("Invalid subcommand.\n")
+		}
+		return true
+	}
+
+	// Parse for bucket and prefix
+	bucket, ok := argv["<bucket>"].(string)
+	if !ok {
```
jonasfj: Required parameters are always present, or docopt will show an error message. No need to do the `ok` part.

s3/s3_test.go
```diff
@@ -0,0 +1,148 @@
...
+
+func TestPutFolder1Folder2(t *testing.T) {
+	filename := createFile()
+	loadConfig()
+
+	// Parse arguments
+	os.Args = []string{"taskcluster", "s3", "put", filename, "test-bucket-for-any-garbage", "folder1/folder2/"}
+	arguments, err := docopt.Parse(usage(), nil, true, version.VersionNumber, true)
+	if err != nil {
+		fmt.Println("Failed to parse arguments")
```
walac: use `assert.NoError`

s3/s3_test.go
```diff
@@ -0,0 +1,148 @@
+package s3
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/docopt/docopt-go"
+	"github.com/stretchr/testify/assert"
```
walac: I would rather prefer [require](https://godoc.org/github.com/stretchr/testify/require):
```go
assert "github.com/stretchr/testify/require"
```

s3/s3_test.go
```diff
@@ -0,0 +1,148 @@
...
+	context.Arguments = arguments
+	result := s3Test.Execute(context)
+
+	assert.Equal(t, true, result, "An error occured during execution.")
+}
+
+func TestGetWithTarget(t *testing.T) {
+	// Get current path
+	pwd, err := os.Getwd()
+	if err != nil {
```
walac: use `assert.NoError`

s3/s3_test.go
```diff
@@ -0,0 +1,148 @@
...
+		fmt.Println("Error occured: ", err)
+	}
+	target := filepath.Join(pwd, "testFile.txt")
+
+	loadConfig()
+
+	// Parse arguments
+	os.Args = []string{"taskcluster", "s3", "get", "test-bucket-for-any-garbage", "folder1/folder2/testFile.txt", target}
+	arguments, err := docopt.Parse(usage(), nil, true, version.VersionNumber, true)
+	if err != nil {
```
walac: use `assert.NoError`

s3/s3_test.go
```diff
@@ -0,0 +1,148 @@
...
+	assert.Equal(t, true, result, "An error occured during execution.")
+}
+
+func TestGetWithTarget(t *testing.T) {
+	// Get current path
+	pwd, err := os.Getwd()
+	if err != nil {
+		fmt.Println("Error occured: ", err)
+	}
+	target := filepath.Join(pwd, "testFile.txt")
```
walac: add `defer os.Remove(target)`

s3/s3_test.go
```diff
@@ -0,0 +1,148 @@
...
+	os.Args = []string{"taskcluster", "s3", "get", "test-bucket-for-any-garbage", "folder1/folder2/testFile.txt", target}
+	arguments, err := docopt.Parse(usage(), nil, true, version.VersionNumber, true)
+	if err != nil {
+		fmt.Println("Failed to parse arguments")
+	}
+
+	context.Arguments = arguments
+	result := s3Test.Execute(context)
+
+	assert.Equal(t, true, result, "An error occured during execution.")
```
walac: nit: add file content assertion to.

s3/s3_test.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+func TestGetWithOutTarget(t *testing.T) {
+	loadConfig()
+
+	// Parse arguments
+	os.Args = []string{"taskcluster", "s3", "get", "test-bucket-for-any-garbage", "folder1/folder2/testFile.txt"}
+	arguments, err := docopt.Parse(usage(), nil, true, version.VersionNumber, true)
+	if err != nil {
```
walac: use `assert.NoError`

s3/s3_test.go
```diff
@@ -0,0 +1,148 @@
...
+		fmt.Println("Failed to parse arguments")
+	}
+
```
walac: add `os.Remove("testFile.txt")`

s3/s3_test.go
```diff
@@ -0,0 +1,148 @@
...
+	assert.Equal(t, true, result, "An error occured during execution.")
+}
+
+func TestGetFolder1(t *testing.T) {
+	loadConfig()
+
+	// Parse arguments
+	os.Args = []string{"taskcluster", "s3", "get", "test-bucket-for-any-garbage", "folder1/testFile.txt"}
+	arguments, err := docopt.Parse(usage(), nil, true, version.VersionNumber, true)
+	if err != nil {
```
walac: use `assert.NoError`

s3/s3_test.go
```diff
@@ -0,0 +1,148 @@
...
+		fmt.Println("Failed to parse arguments")
+	}
+
```
walac: add `os.Remove("testFile.txt")`

s3/s3_test.go
```diff
@@ -0,0 +1,148 @@
...
+	context.Arguments = arguments
+	result := s3Test.Execute(context)
+
+	assert.Equal(t, false, result, "Expected an error.")
+}
+
+func TestPutToFile(t *testing.T) {
+	filename := createFile()
```
walac: add `os.Remove(filename)`

s3/s3_test.go
```diff
@@ -0,0 +1,148 @@
...
+	loadConfig()
+
+	// Parse arguments
+	os.Args = []string{"taskcluster", "s3", "put", filename, "test-bucket-for-any-garbage", "folder1/folder2/testFile.txt"}
+	arguments, err := docopt.Parse(usage(), nil, true, version.VersionNumber, true)
+	if err != nil {
```
walac: use `assert.NoError`

nikinath:
I believe the panic occurs while running go test because the credentials are set to nil as seen [here](https://github.com/taskcluster/taskcluster-cli/blob/master/config/config.go#L55). I could reproduce it by setting the certificate to nil. 
It passes for me if I sign in using `taskcluster signin` command. I might be way off-base but does it make sense to set the default value as `""`? This way it causes a 401 error and not a panic.

I'll make a commit with the necessary changes soon. Also, thank you for your feedback! I appreciate it. I am sorry this PR is taking so long to get resolved. It shouldn't have to be like this. :(

walac:
@jonasfj is the certificate really necessary?
@nikinath no worries, long iterations are expected for any non trivial PR :)

jonasfj:
Certificate is present for temporary credentials, for permanent credentials there is no certificate...

Instead of doing `Certificate: config["config"]["certificate"].(string)`
We should probably do something like: `Certificate: cert` where:
```go
cert, _  := config["config"]["certificate"].(string)
```

That way, if the `.(string)` casting fails we get an empty string instead of a panic. Empty string is also the default value of `Certificate` property, so that's perfect.

djmitche:

PR #34, opened 2016-10-09
Added expand scopes command

### LINKED ISSUE #31: Add expand-scopes command
In taskcluster we use scopes for authorization, see:
https://docs.taskcluster.net/presentations/scopes/
(formal definition and examples)

The taskcluster-auth service maintains roles... a role is effectively just a mapping from a scope on the form `assume:...` to a set of scopes that the role grants.

Example:
`assume:project:taskcluster:tutorial` grants scopes listed here:
https://tools.taskcluster.net/auth/roles/#project:taskcluster:tutorial

A command like this:

``` sh
$ taskcluster expand-scopes assume:project:taskcluster:tutorial
assume:project:taskcluster:tutorial
queue:create-task:aws-provisioner-v1/tutorial
secrets:get:garbage/*
secrets:set:garbage/*
```

Would be pretty nice to have. The API has a method for expanding scopes here:
https://docs.taskcluster.net/reference/platform/auth/api-docs#expandScopes

So this is really just parsing arguments, call API and print the result line by line.

nikinath:
`taskcluster scope` is the main command and `taskcluster scope expand <scope>...` is the subcommand.

Fixes #31.

/cc @walac for review.


scope/scope.go
```diff
@@ -0,0 +1,76 @@
...
+	return nil
+}
+
+func (scope) Summary() string {
+	return "Expand the given scope set."
+}
+
+func usage() string {
+	return `Usage:
+  taskcluster SCOPE expand <scope>...
```
jonasfj: What is the `SCOPE` parameter?

scope/scope.go
```diff
@@ -0,0 +1,76 @@
...
+
+func (scope) Usage() string {
+	return usage()
+}
+
+func (scope) Execute(context extpoints.Context) bool {
+	argv := context.Arguments
+
+	command := argv["SCOPE"].(string)
+	provider := extpoints.CommandProviders()[command]
```
jonasfj: This makes no sense to me... is there is a copy/paste thing going wrong here.

scope/scope.go
```diff
@@ -0,0 +1,76 @@
...
+	if provider == nil {
+		panic(fmt.Sprintf("Unknown command: %s", command))
+	}
+
+	// Set credentials
+	authCreds := tcclient.Credentials(*context.Credentials)
+	myAuth := auth.New(&authCreds)
```
jonasfj: Checkout: https://docs.taskcluster.net/reference/platform/auth/api-docs#expandScopes

No scopes, and by implication no credentials are required. I'm sure you can just create empty Auth with empty credentials and set Authenticate: false:
https://github.com/taskcluster/taskcluster-client-go/blob/master/auth/auth.go#L92

scope/scope.go
```diff
@@ -0,0 +1,70 @@
...
+import (
+	"fmt"
+
+	"github.com/taskcluster/taskcluster-cli/extpoints"
+	"github.com/taskcluster/taskcluster-client-go"
+	"github.com/taskcluster/taskcluster-client-go/auth"
+)
+
+func init() {
+	extpoints.Register("scope", scope{})
```
jonasfj: probably `expand-scope`

scope/scope.go
```diff
@@ -0,0 +1,70 @@
...
+	return nil
+}
+
+func (scope) Summary() string {
+	return "Expand the given scope set."
+}
+
+func usage() string {
+	return `Usage:
+  taskcluster scope expand <scope>...
```
jonasfj: `expand-scope`

scope/scope.go
jonasfj: as we probably will add other commands called something with scope...

scope/scope.go
```diff
@@ -0,0 +1,70 @@
...
+
+func (scope) Usage() string {
+	return usage()
+}
+
+func (scope) Execute(context extpoints.Context) bool {
+	argv := context.Arguments
+
+	authCreds := tcclient.Credentials{}
+	myAuth := auth.New(&authCreds)
```
jonasfj: nit:
`a := auth.New(&tcclient.Credentials{})`

And let's not pass it as argument to `expandScope` we can just create it inside `expandScope`..

scope/scope.go
```diff
@@ -0,0 +1,70 @@
...
+	myAuth.Authenticate = false
+
+	if argv["expand"].(bool) {
+		return expandScope(argv, myAuth)
+	}
+	return true
+}
+
+func expandScope(argv map[string]interface{}, myAuth *auth.Auth) bool {
```
jonasfj: nit... but ideally we do the argument casting in `Execute` and just take `inputScopes []string` as argument here...

nikinath:
Should I add a test case for this? 
For this, I'll have to hardcore the expansion of a scope in the test file - is that okay? Are the scopes subject to change in the future?

jonasfj:
hmm... could we also make the `baseUrl` configurable and then in tests run a test server using `httptest`.

Or we find something that is likely to not change...
Example: `assume:repo:hg.mozilla.org/try:*` expands to something that includes `assume:moz-tree:level:1`, and not assert on the other scopes in the expansion...

nikinath:
It was discussed before to use [require](https://godoc.org/github.com/stretchr/testify/require) instead of [assert](https://godoc.org/github.com/stretchr/testify/assert) and to use `assert.NoError`. Unfortunately, `require` has some known [bugs](https://github.com/stretchr/testify/issues/368) which were proving it difficult to use. Hence, I have used `assert` for now.

expandscope/expandscope.go
```diff
@@ -0,0 +1,67 @@
...
+	a := auth.New(&tcclient.Credentials{})
+	a.Authenticate = false
+
+	params := &auth.SetOfScopes{
+		Scopes: inputScopes,
+	}
+
+	resp, err := a.ExpandScopes(params)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error expanding scopes: %s\n", err)
```
walac: we should return an error here, shouldn't we? Is there a test case for that?

djmitche:

PR #35, opened 2016-10-11
added a scope-check command

nanjekyejoannah:
# I have added a scope-check command.

### DIFF
--- a/scopecheck/scopecheck.go
+++ b/scopecheck/scopecheck.go
@@ -1,0 +1,99 @@
+package scopecheck
+
+import (
+	"fmt"
+
+	"github.com/taskcluster/taskcluster-cli/extpoints"
+	tcclient "github.com/taskcluster/taskcluster-client-go"
+	scopeutility "github.com/taskcluster/taskcluster-lib-scopes"
+)
+
+func init() {
+	extpoints.Register("scope-check", scopecheck{})
+}
+
+type scopecheck struct{}
+
+func (scopecheck) ConfigOptions() map[string]extpoints.ConfigOption {
+	return nil
+}
+
+func (scopecheck) Summary() string {
+	return "Shows whether a given scope satifies another."
+}
+
+func usage() string {
+	return `Usage:
+  taskcluster scope-check <scope1> satifies <scope2>
+`
+}
+
+func (scopecheck) Usage() string {
+	return usage()
+}
+
+type Auth tcclient.ConnectionData
+
+func New(credentials *tcclient.Credentials) *Auth {
+	myAuth := Auth(tcclient.ConnectionData{
+		Credentials:  credentials,
+		BaseURL:      "https://auth.taskcluster.net/v1",
+		Authenticate: true,
+	})
+	return &myAuth
+}
+
+//may have duplicate of this method because this method will be written at some point in issue#31
+func (myAuth *Auth) expandScope(payload *SetOfScopes) (*SetOfScopes, error) {
+	connectionDetails := tcclient.ConnectionData(*myAuth)
+	responseObject, _, err := (&connectionDetails).APICall(payload, "GET", "/scopes/expand", new(SetOfScopes), nil)
+	return responseObject.(*SetOfScopes), err
+}
+
+func (scopecheck) Execute(context extpoints.Context) bool {
+	argv := context.Arguments
+
+	command := argv["scope-check"].(string)
+	provider := extpoints.CommandProviders()[command]
+	if provider == nil {
+		panic(fmt.Sprintf("Unknown command: %s", command))
+		return false
+	}
+
+	satisfies := argv("<satisfies>")
+	rscope := argv("<rscope>")
+	lscope := argv("<lscope>")
+
+	if argv["rscope"].(bool) {
+
+		response := checkscopes(rscope, lscope)
+		fmt.Printf("%s\n", response)
+	}
+	return true
+}
+
+func checkscopes(rightScope *SetOfScopes, leftScope *SetOfScopes) string {
+
+	if rightScope == leftScope {
+
+		return "YES"
+	} else {
+
+		scope, errs := expandScope(leftScope)
+		if errs != nil {
+			resp := "Error while trying to expand scopes"
+			return resp
+		}
+
+		if scopeutility.scopeMatch(rightScope, scope) {
+			resp := "YES"
+			return resp
+		} else {
+			resp := "NO missing -"
+			resp += leftScope
+			return resp
+		}
+
+	}
+
+}
--- a/scopecheck/scopecheck.go
+++ b/scopecheck/scopecheck.go
@@ -50,23 +50,23 @@ func expandScope(scope2 []string) (*auth.SetOfScopes, error) {
 func (scopecheck) Execute(context extpoints.Context) bool {
 	argv := context.Arguments
 	scope1 := argv["<scope1>"].(string)
-	if argv["satisfies"] == true {
-		scope2 := argv["<scope2>"].(string)
-		response := checkscopes(scope1, scope2)
-		fmt.Printf("%s\n", response)
-	}
-
+	satisfies := argv["satisfies"]
+	scope2 := argv["<scope2>"].(string)
+	response := checkscopes(scope1, scope2)
+	fmt.Printf("%s\n", response)
 	return true
 }
 
 func checkscopes(scope1 string, scope2 string) string {
 	var secondScope interface{} = scope2
 	if strings.HasPrefix("scope2", "assume:") {
+		//cast scope2 to string array before calling api method expandingScope
 		expandedScope, errs := expandScope(secondScope.([]string))
 		if errs != nil {
 			resp := "Error while trying to expand scopes"
 			return resp
 		}
+		//Need it back to string for comparison with scope1
 		var expdscope interface{} = expandedScope
 		scope2 = expdscope.(string)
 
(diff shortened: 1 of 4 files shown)


scopecheck/test_scopecheck.go
```diff
@@ -0,0 +1,19 @@
+package scopecheck
```
jonasfj: test cases are awesome, let's name it `scopecheck_test.go` then it only gets compiled when running `go test ./...`

subtree_imports.go
```diff
@@ -3,3 +3,5 @@ package main
 
 import _ "github.com/taskcluster/taskcluster-cli/apis"
 import _ "github.com/taskcluster/taskcluster-cli/extpoints"
+import _ "github.com/taskcluster/taskcluster-cli/scopecheck"
+import _ "github.com/taskcluster/taskcluster-cli/scope"
```
jonasfj: probably a line too much here...

scopecheck/scopecheck.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+func usage() string {
+	return `Usage:
+  taskcluster scope-check <scope1> satifies <scope2>
+`
+}
+
+func (scopecheck) Usage() string {
+	return usage()
```
jonasfj: Why the helper method?

scopecheck/scopecheck.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+type Auth tcclient.ConnectionData
+
+func New(credentials *tcclient.Credentials) *Auth {
```
jonasfj: Why export this function? does it even need to be a function?

scopecheck/scopecheck.go
```diff
@@ -0,0 +1,99 @@
...
+	myAuth := Auth(tcclient.ConnectionData{
+		Credentials:  credentials,
+		BaseURL:      "https://auth.taskcluster.net/v1",
```
jonasfj: this has sane defaults...

scopecheck/scopecheck.go
```diff
@@ -0,0 +1,99 @@
...
+		Authenticate: true,
+	})
+	return &myAuth
+}
+
+//may have duplicate of this method because this method will be written at some point in issue#31
+func (myAuth *Auth) expandScope(payload *SetOfScopes) (*SetOfScopes, error) {
```
jonasfj: Why not use:
https://godoc.org/github.com/taskcluster/taskcluster-client-go/auth#Auth.ExpandScopes

scopecheck/scopecheck.go
```diff
@@ -0,0 +1,99 @@
...
+	connectionDetails := tcclient.ConnectionData(*myAuth)
+	responseObject, _, err := (&connectionDetails).APICall(payload, "GET", "/scopes/expand", new(SetOfScopes), nil)
+	return responseObject.(*SetOfScopes), err
+}
+
+func (scopecheck) Execute(context extpoints.Context) bool {
+	argv := context.Arguments
+
+	command := argv["scope-check"].(string)
+	provider := extpoints.CommandProviders()[command]
```
jonasfj: I'm confused?

scopecheck/scopecheck.go
```diff
@@ -0,0 +1,99 @@
...
+	if provider == nil {
+		panic(fmt.Sprintf("Unknown command: %s", command))
+		return false
+	}
+
+	satisfies := argv("<satisfies>")
+	rscope := argv("<rscope>")
+	lscope := argv("<lscope>")
```
jonasfj: scope1 and scope2

scopecheck/scopecheck.go
```diff
@@ -0,0 +1,94 @@
...
+
+	_, expandedScopes := response.Scopes
+	return expandedScopes.(string), nil
+}
+
+func (scopecheck) Execute(context extpoints.Context) bool {
+	argv := context.Arguments
+
+	command := argv["SCOPECHECK"].(string)
+	provider := extpoints.CommandProviders()[command]
```
jonasfj: Still no clue what this is good for... Why are you touching CommandProviders()?

What is `argv["SCOPECHECK"].(string)`...

Please take some time and make sure you understand the code and why you have it. Because this doesn't seem to make sense.

scopecheck/scopecheck.go
```diff
@@ -0,0 +1,94 @@
...
+	if provider == nil {
+		panic(fmt.Sprintf("Unknown command: %s", command))
+		return false
+	}
+
+	satisfies := argv("<satisfies>")
```
jonasfj: `argv isn't a function... this won't build, and again, please make sure you understand what the code is supposed to do.

scopecheck/scopecheck.go
```diff
@@ -0,0 +1,87 @@
...
+
+	return resp, err
+}
+
+func (scopecheck) Execute(context extpoints.Context) bool {
+	argv := context.Arguments
+	scope1 := argv["<scope1>"].(string)
+	scope2 := argv["<scope2>"].(string)
+
+	if argv["scope-check"].(bool) {
```
jonasfj: What is this check good for?

nanjekyejoannah: After thinking about it..this looks irrelevant...

nanjekyejoannah:
@jonasfj  can you give me a hint on where I may be going wrong and what I should do?

scopecheck/scopecheck.go
jonasfj: Yes

subtree_imports.go
```diff
@@ -3,3 +3,5 @@ package main
 
 import _ "github.com/taskcluster/taskcluster-cli/apis"
 import _ "github.com/taskcluster/taskcluster-cli/extpoints"
+import _ "github.com/taskcluster/taskcluster-cli/scopecheck"
+import _ "github.com/taskcluster/taskcluster-client-go/auth"
```
jonasfj: Any idea how that line ended up there?

scopecheck/scopecheck.go
```diff
@@ -0,0 +1,85 @@
...
+		return nil, err
+	}
+
+	return resp, err
+}
+
+func (scopecheck) Execute(context extpoints.Context) bool {
+	argv := context.Arguments
+	scope1 := argv["<scope1>"].(string)
+	if argv["satisfies"] == true {
```
jonasfj: What is the point of this check?

nanjekyejoannah: For the satisfies keyword.

nanjekyejoannah: If I just create a variable for this keyword .. I will end up not using it anywhere and I will get a compile error for variable not being used...

jonasfj: Well, do you need to create a variable?
It's just a syntax separating scope1 and scope2 right... Why would you need to check that it's there...

nanjekyejoannah: @jonasfj Well to enable use of syntax of a form like `taskcluster scope-check scope1 satisfies scope2` I will introduce logic of this form `satisfies =: argv["<scope1>"].(string)` but I will not use the satisfies variable anywhere in my remaining logic it is to just enable that syntax. so the compiler complains(variable that is not used) but if I just performed this `argv["satisfies"] == true` I avail the syntax and make the compiler happy.....could there be a cleaner way of me achieving this..than this way?..by the way thanks for the review and sorry for the back and forth...

jonasfj: The arguments have already been parsed when you do: `satisfies =: argv["satisfies"]`, so how could it possibly affect the argument-parsing...

Arguments are parsed based on the `Usage()` string and passed into `Execute(context extpoints.Context)` as a property on context... Consider:
https://github.com/taskcluster/taskcluster-cli/blob/master/main.go#L88-L114

nanjekyejoannah: when I do `satisfies =: argv["satisfies"]` I am introducing a variable satisfies which the compiler will complain about because I wont use it anywhere later...

scopecheck/scopecheck.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	return resp, err
+}
+
+func (scopecheck) Execute(context extpoints.Context) bool {
+	argv := context.Arguments
+	scope1 := argv["<scope1>"].(string)
+	satisfies := argv["satisfies"]
+	_ = satisfies
```
walac: ?

scopecheck/scopecheck.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+func checkscopes(scope1 string, scope2 string) string {
+	var secondScope interface{} = scope2
+	if strings.HasPrefix("scope2", "assume:") {
+		//cast scope2 to string array before calling api method expandingScope
+		expandedScope, errs := expandScope(secondScope.([]string))
+		if errs != nil {
+			resp := "Error while trying to expand scopes"
+			return resp
```
walac: why not just
```go
return "Error while trying to expand scopes"
```
?

scopecheck/scopecheck.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	if scope1 == scope2 {
+		return "YES"
+	} else {
+		resp := "NO missing -"
+		return resp
+	}
+
+}
```
walac: I think of all this might be replaced by [this](https://godoc.org/github.com/taskcluster/taskcluster-base-go/scopes)

djmitche:

PR #43, opened 2017-01-05
Running any api command creates `-` in the working directory

djmitche:
```
154     var output io.Writer
155     if out, ok := argv["--output"].(string); ok {
156         f, err := os.Create(out)
```
where the default for `output` is, I assume, `-`.

### DIFF
--- a/apis/provider.go
+++ b/apis/provider.go
@@ -151,15 +151,19 @@ func (p apiProvider) Execute(context extpoints.Context) bool {
 	// Setup output
 	var output io.Writer
 	if out, ok := argv["--output"].(string); ok {
-		f, err := os.Create(out)
-		if err != nil {
-			fmt.Printf("Failed to open output file, error: %s\n", err)
-			return false
+		if out == "-" {
+			output = os.Stdout
+		} else {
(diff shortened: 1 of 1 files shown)


apis/provider.go
```diff
@@ -151,15 +151,19 @@ func (p apiProvider) Execute(context extpoints.Context) bool {
...
+				return false
+			}
+			defer f.Close()
+			output = f
 		}
-		defer f.Close()
-		output = f
 	} else {
-		output = os.Stdout
+		panic("cannot determine --output")
```
jonasfj: Shouldn't this just not be a case? Given that `--output` has a default value it can't not have a value right?

jonasfj: This feels more like a case that isn't necessary... IMO, if we default output to stdout, then we don't have to do this...

jonasfj: I would just do:
```go
var output io.Writer = os.Stdout
if out := argv["--output"].(string); out != "-" {
  f, err := os.Create(out)
  if err != nil {
    fmt.Printf(...)
    return false
  }
  defer f.Close()
  output = f
}
```
This is all nit...

Issue #20, opened 2016-09-15
Redirect error messages to stderr


walac:
Currently, all errors message go to `os.Stdout`. We should use `fmt.Fprintf` to redirect them to `os.Stderr`.

ApheleiaS:
Hello, 
Please assign me to this bug. I am interested in participating in outreachy this round. 
Thanks.

iamneha:
@ApheleiaS I almost solved this bug I would like to request you to please ask for another bug and I am sorry that I forgot to ask mentor to assign this to me.

@walac can you assign me this bug

ApheleiaS:
@iamneha  I solved this bug quite a bit as well. So I am not sure.  Anyhow you should make your PR if you are done. That way we can learn from each others code and devs can merge the one which looks better :)

walac:
@iamneha sorry, but @ApheleiaS requested bug assignment first. We are working this week to get more good-first-bugs, as the number of applicants is way beyond our expectations.

yannlandry:
Started working on this issue. Seems like it's gonna be a tedious task of search and replace - wish me luck.

PR #69, opened 2017-01-20
Send errors to stderr instead of stdout (issue #20)

yannlandry:
Fixes issue #20, went with the simple solution of Fprintf everywhere.


AyubMohamed:
Looks good, there seem to be a few logs that still use `log.`Println("", err)` with an error in some test files such as _example_test.go_ and _retry_test.go_ to name a few. However I don't know if we care about these haha

PR #70, opened 2017-01-20
add LICENSE

djmitche:
Jonas, I think you're the original author.  MPLv2 OK?


djmitche:
@t0xicCode fyi, MPLv2 it is

PR #73, opened 2017-01-21
Add from-now cli command

AyubMohamed:


from_now/from_now.go
```diff
@@ -0,0 +1,112 @@
...
+	return usage
+}
+
+func (from_now) Execute(context extpoints.Context) bool {
+	duration := context.Arguments["DURATION"].(string)
+
+	offset, err := parseTime(duration)
+	
+	if err {
+		panic("String: '" + duration + "' isn't a time expression")
```
yannlandry: The convention here seems to be to output to `os.Stderr` and `return false`. We tend to reserve `panic` for when things go really bad (a segfault, for example).

    fmt.Fprintf(os.Stderr, "Formatted error message")
    return false

from_now/from_now.go
```diff
@@ -0,0 +1,112 @@
...
+		"(\\s*(\\d+)\\s*h((ours?)|r)?)?",
+		"(\\s*(\\d+)\\s*min(utes?)?)?",
+		"(\\s*(\\d+)\\s*s(ec(onds?)?)?)?",
+		"\\s*$",
+	}
+
+	re := regexp.MustCompile(strings.Join(reg, ""))
+
+	if !re.MatchString(str) {
+		return parse_time{}, true
```
yannlandry: Is that what happens when `str` is not valid? If that's what it is, I would suggest returning `nil` to indicate that parsing failed.

from_now/from_now.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+/* parseTime takes an argument `str` which is a string on the form `1 day 2 hours 3 minutes`
+ * where specification of day, hours and minutes is optional. You can also the
+ * short hand `1d2h3min`, it's fairly tolerant of different spelling forms and
+ * whitespace. But only really meant to be used with constants.
+ *
+ * Returns a parse_time object with all of the fields filled in with the correct values.
+ */
+func parseTime(str string) (parse_time, bool) {
```
yannlandry: Does the `bool` return type indicate that the parsing failed? I would recommend using [Go's own error type](https://blog.golang.org/error-handling-and-go). You can do `return nil, errors.New("Your error message")` in case of failure.

from_now/from_now.go
```diff
@@ -0,0 +1,112 @@
...
+
+	if !re.MatchString(str) {
+		return parse_time{}, true
+	}
+
+	groupMatches := re.FindAllStringSubmatch(str, -1)
+
+	offset := parse_time{}
+
+	// Add negative support after we figure out what we are doing with docopt because it complains about the '-'
```
yannlandry: Idea: add `minus` command or `--negative` option.

from_now/from_now_test.go
```diff
@@ -0,0 +1,42 @@
...
+		t.Error("Expected 6 minutes, got ", offset.minutes)
+	}
+
+	if offset.seconds != 7 {
+		t.Error("Expected 7 seconds, got ", offset.seconds)
+	}
+
+	// Test if it's a valid time expression.
+	offset, err = parseTime("this should produce an error.")
+	if !err {
```
yannlandry: You'll probably need to change this if you use Go's errors. Usually `if err != nil`.

version/version.go
```diff
@@ -33,12 +33,6 @@ func (version) Usage() string {
 }
 
 func (version) Execute(context extpoints.Context) bool {
```
yannlandry: I would have sent this change under a different PR/code review since it's a different feature, but otherwise there is no issue here.

djmitche: This is a good point, although the fix is nicely isolated in its own commit, which is often enough for something fairly simple.

from_now/from_now.go
AyubMohamed: I think once waiting to figure out what we're doing with _docopt_ makes sense here because if we scrap it adding the minus command would be useless.

from_now/from_now.go
```diff
@@ -0,0 +1,115 @@
...
+import (
+	"fmt"
+	"github.com/taskcluster/taskcluster-cli/extpoints"
+	"regexp"
+	"strconv"
+	"strings"
+	"time"
+	"os"
+	"errors"
+)
```
walac: Use [goimports](https://godoc.org/golang.org/x/tools/cmd/goimports) as coding format tool.

from_now/from_now.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+func (from_now) Summary() string {
+	return "Returns a timestamp which is <duration> ahead in the future."
+}
+
+func (from_now) Usage() string {
+	usage := "Usage: taskcluster from-now <duration>"
+	usage += "\n"
+	return usage
```
walac: Why not just:

```go
return "Usage: taskcluster from-now <duration>\n"
```
?

from_now/from_now.go
```diff
@@ -0,0 +1,115 @@
...
+	years   int
+	months  int
+	weeks   int
+	days    int
+	hours   int
+	minutes int
+	seconds int
+}
+
+/* parseTime takes an argument `str` which is a string on the form `1 day 2 hours 3 minutes`
```
walac: nit: comment
```go
/*
 * parseTime ...
 */
```

from_now/from_now.go
```diff
@@ -0,0 +1,115 @@
...
+ * where specification of day, hours and minutes is optional. You can also the
```
walac: nit: "You can also short the short hand"? I think sentence is incomplete...

from_now/from_now.go
```diff
@@ -0,0 +1,115 @@
...
+	// 	neg = -1
+	// }
+
+	offset.years, _ = strconv.Atoi(groupMatches[0][4]) 
+	offset.months, _ = strconv.Atoi(groupMatches[0][8]) 
+	offset.weeks, _ = strconv.Atoi(groupMatches[0][11]) 
+	offset.days, _ = strconv.Atoi(groupMatches[0][15]) 
+	offset.hours, _ = strconv.Atoi(groupMatches[0][18]) 
+	offset.minutes, _ = strconv.Atoi(groupMatches[0][22]) 
+	offset.seconds, _ = strconv.Atoi(groupMatches[0][25])
```
walac: I am not sure if ignoring errors at all is a good idea. How about a helper function that [panic](https://blog.golang.org/defer-panic-and-recover) in case of error?

from_now/from_now_test.go
```diff
@@ -0,0 +1,42 @@
+package from_now
+
+import "testing"
+
+func TestparseTime(t *testing.T) {
+
```
walac: nit: remove blank line.

from_now/from_now_test.go
```diff
@@ -0,0 +1,42 @@
...
+	if offset.seconds != 7 {
+		t.Error("Expected 7 seconds, got ", offset.seconds)
+	}
+
+	// Test if it's a valid time expression.
+	offset, err = parseTime("this should produce an error.")
+	if err != nil {
+		t.Error("Expected an error, got ", offset)
+	}
+}
```
walac: nit: use [testify](https://github.com/stretchr/testify) for test assertions.

from_now/from_now_test.go
```diff
@@ -0,0 +1,23 @@
+package from_now
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
```
walac: Use:
```go
assert "github.com/stretchr/testify/require"
```

`require` will fail the test in case of assertion violation, whereas `assert` just return a boolean to indicating if assertion failed or not.

PR #67, opened 2017-01-21
Format the API references readably

t0xicCode:
Right now the generated API references are a single very long line.

Let's find a way to make that readable, so that diffs make sense ("oh, this API method was added").


apis/_codegen/fetch-apis.go
```diff
@@ -1,99 +1,235 @@
...
+			g.Printf("(nil)")
+			break
+		}
+		g.Print("{\n")
+		if sorted && t.Elem().Kind() == reflect.String {
+			sortedK := make([]string, 0, v.Len())
+			for i := 0; i < v.Len(); i++ {
+				sortedK = append(sortedK, v.Index(i).String())
+			}
+			sort.Strings(sortedK)
```
djmitche: This is a little disconcerting, sorting arrays of strings.  From what I can tell, this ordering doesn't matter in the data being handled here, but I'd like to see a warning comment about it!  @jonasfj does this seem OK to you?

apis/_codegen/fetch-apis.go
```diff
@@ -1,99 +1,235 @@
...
+	gen.PrintComposite(schemas, true)
+	gen.Print("\n")
 
-	source, err := format.Source([]byte(code))
+	// Format the output.
+	source, err := gen.Format()
 	if err != nil {
-		fmt.Println("go fmt, code generation failed, error: ", err)
-		os.Exit(1)
+		log.Println(gen.String())
```
yannlandry: ...or does that print the code generated until now?

apis/_codegen/fetch-apis.go
```diff
@@ -1,99 +1,235 @@
...
+func (g *generator) Printf(format string, args ...interface{}) {
+	fmt.Fprintf(&g.buf, format, args...)
+}
+
+// Print prints the given a to the buffer.
+func (g *generator) Print(a ...interface{}) {
+	fmt.Fprint(&g.buf, a...)
+}
+
+func (g *generator) PrintComposite(data interface{}, sorted bool) {
```
djmitche: It looks like `sorted` is always true.  Perhaps it could be omitted?

apis/_codegen/fetch-apis.go
djmitche: This also assumes that arrays of non-strings need not be sorted, and thus that their order is stable.  Is that assumption invalid for arrays of strings?

apis/_codegen/fetch-apis.go
```diff
@@ -1,99 +1,235 @@
...
+		g.Print("}")
+	case reflect.Map:
+		// finally map
+		g.Printf("%s{\n", t.String())
+		keys := v.MapKeys()
+		if len(keys) == 0 {
+			g.Print("}")
+			break
+		}
+		if sorted && t.Key().Kind() == reflect.String {
```
djmitche: I would argue that sorting always makes sense here, regardless of `sorted`.  The nondeterminism we want to avoid is in map ordering, rather than in array order in upstream schema/reference generation.

apis/_codegen/fetch-apis.go
jonasfj: Yeah, doc comments are useful :)

apis/_codegen/fetch-apis.go
jonasfj: I'm not sure what you're doing here... but  JSON always becomes `Array` or `Slice` of type `interface{}` so if this runs of JSON it'll never get called...

If trying to pretty print JSON this will never run... If you were pretty printing JSON, I wouldn't sort strings as their ordering might carry semantics.

jonasfj: Please add a doc comment for the method explaining briefly what it does.

apis/_codegen/fetch-apis.go
t0xicCode: It does. Under normal circumstances that line should not be executed because we've ensured that the output we generate is valid go code. However it is useful for debugging the code when we make changes. I'll take it out.

apis/_codegen/fetch-apis.go
```diff
@@ -1,99 +1,235 @@
...
-		// Parse reference
-		var s definitions.Service
-		if err2 := json.Unmarshal(res.Body, &s); err2 != nil {
-			fmt.Println("Failed parse API: ", name, " error: ", err2)
-			os.Exit(1)
-		}
-		services[name] = s
+		wg.Add(1)
+		go func(n string, u string) {
+			s := fetchService(g, n, u)
```
t0xicCode: I did, I read through go-got and didn't see anything that would be thread-unsafe (no writes to shared objects/memory). I'll add a comment though.

apis/_codegen/fetch-apis.go
t0xicCode: 👍

t0xicCode: sure thing. I was thinking that eventually this method might be extracted into a separate package, but we can cross that bridge when/if we get there.

djmitche:
Looks like this is still awaiting a second review by @yannlandry and a first by @AyubMohamed

yannlandry:
Looks like there won't be anymore reviews on here. Let's merge.

PR #78, opened 2017-01-21
Authentication for API commands doesn't match the server's expected format

t0xicCode:
Any command that uses the auto-generated API commands with authentication will not work because the `ext` parameter of the hawk `Authorization` header is not base64 encoded.

### DIFF
--- a/client/credentials.go
+++ b/client/credentials.go
@@ -82,11 +82,7 @@ func (c *Credentials) newAuth(method, url string, h hash.Hash) (*hawk.Auth, erro
 	}
 	if e.Certificate != nil || e.AuthorizedScopes != nil {
 		s, _ := json.Marshal(e)
-		if string(s) != "{}" {
-			a.Ext = base64.StdEncoding.EncodeToString(s)
-		} else {
-			a.Ext = string(s)
-		}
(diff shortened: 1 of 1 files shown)


djmitche:
@petemoore I feel like this must have worked once, so while this fix looks correct, and works .. why are we only noticing this now?  Any ideas?

petemoore:
This was @jonasfj's baby - I never used taskcluster-cli, so I couldn't say if it was working at some point or not. :/

@jonasfj - do you know?

jonasfj:
It probably never worked :)

client/credentials.go
```diff
@@ -82,7 +82,11 @@ func (c *Credentials) newAuth(method, url string, h hash.Hash) (*hawk.Auth, erro
 	}
 	if e.Certificate != nil || e.AuthorizedScopes != nil {
 		s, _ := json.Marshal(e)
-		a.Ext = string(s)
+		if string(s) != "{}" {
```
jonasfj: `string(s)` can't be `{}` if `e.Certificate != nil || e.AuthorizedScopes != nil`, or am I missing something?

PR #23, opened 2017-01-23
Add a `task` subcommand for simple task queries

t0xicCode:
Summary:

``` shell
taskcluster task status <taskId>
```

### DIFF
--- a/task/task.go
+++ b/task/task.go
@@ -27,6 +27,7 @@ func (task) Usage() string {
 Usage:
   taskcluster task status [--all-runs] [--] <taskId>
   taskcluster task name [--] <taskId>
+  taskcluster task group [--] <taskId>
 
 Options:
   --all-runs  Use all runs instead of only the latest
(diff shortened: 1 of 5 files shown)


djmitche:
Note that the referenced "download command" is something @jhford is working
on.

task/task.go
```diff
@@ -0,0 +1,78 @@
...
+		return executeSubCommand(context, t.runComplete)
+	}
+
+	return false
+}
+
+// executeSubCommand executes the given SubCommand.
+func executeSubCommand(context extpoints.Context, subCommand SubCommand) bool {
+	var c *tcclient.Credentials
+	if context.Credentials != nil {
```
t0xicCode: We shouldn't because `ToClientCredentials` accesses members of the `Credentials` struct.

In go, it is idiomatic to have the caller check for nil or other erroneous state and let the called function assume its receiver is not nil (so it can access fields and what not).

PR #80, opened 2017-01-23
Fixed linting errors with the from_now package

t0xicCode:
The from_now package as committed had a few linting issues. Namely regarding the use of the `_` character. This PR fixes these issues by renaming the symbols to match the go idioms: `from_now -> from-now|fromNow`.

As an added bonus, I've added a few tests, improving our code-coverage in this package to a nice 60%.


fromNow/from_now_test.go
```diff
@@ -0,0 +1,57 @@
...
+
+	assert "github.com/stretchr/testify/require"
+)
+
+func TestParseTimeComplete(t *testing.T) {
+	assert := assert.New(t)
+	offset, err := parseTime("1 year 2 months 3 weeks 4 days 5 hours 6 minutes 7 seconds")
+
+	assert.NoError(err, "err should be nil")
+	assert.Equal(offset.years, 1, "they should be equal")
```
yannlandry: I know it's not originally your code, but I think these error messages aren't very descriptive. Can we have something that tells me more than simply "they should be equal"?

t0xicCode:
Sure thing.

PR #82, opened 2017-01-27
Regexp tweaks for slugid library

petemoore:
This tightens the regexp checks for slugs and uuids. Furthermore,
we now compile the regexp expressions upfront, to avoid compiling
with every decode/encode call.

Lastly, adjusted the tests for generating slugs, to generate 1000
slugs rather than one, to increase confidence in results. This
added 4ms to the test time, which seems a reasonable cost/benefit.


slugid/slugid.go
```diff
@@ -10,6 +10,26 @@ import (
...
+	// strings.
+
+	// V4_SLUG_REGEXP is the regular expression that all V4 Slug IDs should conform to
+	V4_SLUG_REGEXP = regexp.MustCompile("^[A-Za-z0-9_-]{8}[Q-T][A-Za-z0-9_-][CGKOSWaeimquy26-][A-Za-z0-9_-]{10}[AQgw]$")
+
+	// V4_UUID_REGEXP is the regular expression that all V4 UUIDs should conform to
+	V4_UUID_REGEXP = regexp.MustCompile("^[a-f0-9]{8}-[a-f0-9]{4}-4[a-f0-9]{3}-[89ab][a-f0-9]{3}-[a-f0-9]{12}$")
+
+	// NICE_SLUG_REGEXP is the regular expression that all "nice" Slug IDs should conform to
+	NICE_SLUG_REGEXP = regexp.MustCompile("^[A-Za-f][A-Za-z0-9_-]{7}[Q-T][A-Za-z0-9_-][CGKOSWaeimquy26-][A-Za-z0-9_-]{10}[AQgw]$")
```
jonasfj: What use does this have in CLI client

slugid/slugid.go
```diff
@@ -77,11 +97,7 @@ func nice() (string, error) {
 func decode(context extpoints.Context) (string, error) {
 	slug := context.Arguments["<slug>"].(string)
 
-	// validation
-	match, err := regexp.MatchString("^[A-Za-z0-9-_]{22}$", slug)
-	if err != nil {
-		return "", fmt.Errorf("Error validating slug: %s", err)
-	}
+	match := V4_SLUG_REGEXP.MatchString(slug)
```
t0xicCode: Would it make sense to add a comment that NICE slugs are a subset of V4 slugs (and thus match the V4 format)?

slugid/slugid.go
```diff
@@ -95,10 +111,7 @@ func encode(context extpoints.Context) (string, error) {
 	uuid := context.Arguments["<uuid>"].(string)
 
 	// validation
-	match, err := regexp.MatchString("^[a-f0-9]{8}-([a-f0-9]{4}-){3}[a-f0-9]{12}$", uuid)
-	if err != nil {
-		return "", fmt.Errorf("Error validating uuid: %s", err)
-	}
+	match := V4_UUID_REGEXP.MatchString(uuid)
```
t0xicCode: Same here: comment that NICE UUIDs are a subset of V4 UUIDs?

slugid/slugid_test.go
```diff
@@ -12,21 +11,23 @@ import (
...
-	slug, _ := v4()
-	match, err := regexp.MatchString("^[A-Za-z0-9-_]{22}$", slug)
-
-	assert.Nil(err, fmt.Sprintf("Error matching string generated by v4(): %s", err))
-	assert.Equal(match, true, fmt.Sprintf("Slug generated by v4() is invalid."))
+	// Let's test 1000 generated slugs, to increase confidence
+	for i := 0; i < 1000; i++ {
+		slug, _ := v4()
+		match := V4_SLUG_REGEXP.MatchString(slug)
+		assert.Equal(match, true, fmt.Sprintf("Slug %q generated by v4() is invalid.", slug))
```
t0xicCode: [`assert.True(match, ...)`](https://godoc.org/github.com/stretchr/testify/require#Assertions.True) can be used instead of `assert.Equal(match, true, ...)`.

slugid/slugid_test.go
```diff
@@ -12,21 +11,23 @@ import (
...
-	slug, _ := nice()
-	match, err := regexp.MatchString("^[A-Za-z0-9-_]{22}$", slug)
-
-	assert.Nil(err, fmt.Sprintf("Error matching string generated by nice(): %s", err))
-	assert.Equal(match, true, fmt.Sprintf("Slug generated by nice() is invalid."))
+	// Let's test 1000 generated slugs, to increase confidence
+	for i := 0; i < 1000; i++ {
+		slug, _ := nice()
+		match := NICE_SLUG_REGEXP.MatchString(slug)
+		assert.Equal(match, true, fmt.Sprintf("Slug %q generated by nice() is invalid.", slug))
```
t0xicCode: Same: [`assert.True(match, ...)`](https://godoc.org/github.com/stretchr/testify/require#Assertions.True) can be used instead of `assert.Equal(match, true, ...)`.

slugid/slugid.go
petemoore: For validating that slugs are compliant.

PR #89, opened 2017-02-01
Add .taskcluster.yml for testing

lteigrob:
This will be a replacement of .travis.yml for testing purposes.

A couple things to note:

The -t flag makes sure all required packages are downloaded for the tests
The /... runs through all subdirectories
(Excluding either of these causes the setup to fail)

The ./... ensures a run-through all available test files. 

The previous merge to change codegen to _codegen was to support use of .taskcluster.yml:  Names starting with an underscore are ignored by 'go test'. We needed codegen to be ignored since it uses the package go-get, which is not part of the default software download so couldn't be found by 'go test', resulting in task failure.


lteigrob:
Third try's a charm. Feel free to laugh at my trailing newline commits. I think she's actually ready to review now. :)

djmitche:
We can squash all of this together when merging :)

.taskcluster.yml
```diff
@@ -0,0 +1,38 @@
...
+  name: "TaskCluster GitHub Tests"
+  description: "All non-integration tests for taskcluster github"
+  owner: "{{ event.head.user.email }}" # the user who sent the pr/push e-mail will be inserted here
+  source: "{{ event.head.repo.url }}"  # the repo where the pr came from will be inserted here
+tasks:
+  # What kind of environment will you need (docker, windows, etc...)
+  - provisionerId: "{{ taskcluster.docker.provisionerId }}"
+    # Be careful with spacing under the lines with dashes. Next line should begin under the word, not under the dash.
+    # Worker types correspond to particular machine types (aws size, etc...)
+    # worker types may be added by priveleged taskcluster users at https://tools.taskcluster.net/aws-provisioner
```
djmitche: Let's delete these comments and the other "how to fill in this file" comments, as they may confuse people later.

djmitche:
Awesome!  Tomorrow's meeting will be a good time to talk about what this does and what can be added to it (not necessarily by you!), so the whole team knows about it.

lteigrob:
Nice, thanks!

PR #90, opened 2017-02-02
Prepare configuration for transition to cobra

yannlandry:
This is mostly refactoring.

I moved the configuration and credentials from the `main()` function to the `config` package and everything gets accessed from there. Now, you simply call `config.Setup()` from `main()`.

A dedicated object to hold `ConfigOption`s has been created, with a function to register new options. It is also wired to integrate with `CommandProvider`s until all the commands have moved to the new process for registering options.

Everything looks good on my end, but since I'm playing in an area of the code I don't know really well you might want to take a closer look by pulling the branch.

This will complete step 1 of the [Transition Plan](https://github.com/taskcluster/taskcluster-cli/wiki/Transition-plan-from-docopt-to-cobra).

### DIFF
--- a/config/init.go
+++ b/config/init.go
@@ -1,0 +1,43 @@
+package config
+
+import (
+    "fmt"
+    "os"
+
+	"github.com/taskcluster/taskcluster-cli/client"
+)
+
+var (
+    Configuration map[string]map[string]interface{}
+    ConfigOptions = make(map[string]map[string]ConfigOption)
+    Credentials *client.Credentials
+)
+
+// Setup: call it from main
+// this was originally the init() function
+// but we want to make sure all other packages have been initialized
+// before calling them, which Load() does
+func Setup() {
+    var err error
+
+    // load configuration
+    Configuration, err = Load()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Failed to load configuration file, error: %s\n", err)
+		os.Exit(1)
+	}
+
+    // load credentials
+    clientID, ok1 := Configuration["config"]["clientId"].(string)
+	accessToken, ok2 := Configuration["config"]["accessToken"].(string)
+	if ok1 && ok2 {
+		certificate, _ := Configuration["config"]["certificate"].(string)
+		authorizedScopes, _ := Configuration["config"]["authorizedScopes"].([]string)
+		Credentials = &client.Credentials{
+			ClientID:         clientID,
+			AccessToken:      accessToken,
+			Certificate:      certificate,
+			AuthorizedScopes: authorizedScopes,
+		}
+	}
+}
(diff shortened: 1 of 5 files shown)


t0xicCode:
nit: could you run `go fmt` on your source?

yannlandry:
Yep, I'll `go fmt` before my next commit.

**As for `ConfigOption`:**

We're going to get rid of `extpoints` at the end of the transition as it's a complement to docopt. I copied `ConfigOption` to the `config` package because it will be used in the future and left the other one in place as it is used by many of the commands and removing it right now would break a lot of things.

Do you know if there's a way to do a `typedef` kind of thing in Go where `extpoints.ConfigOption` would be a 'reference' to `config.ConfigOption`.

t0xicCode:
There is a way, it's called struct composition or struct nesting, but you
can also use gofmt (not go fmt) to programmatically rewrite all references
of extpoints.ConfigOption into references to config.ConfigOption.

That way the change is entirely contained in the commit.

yannlandry:
I've looked into this and there is no quick way to make a reference (or struct nesting, or typedef) from `extpoints.ConfigOption` to `config.ConfigOption`. Since `extpoints` calls its own implementation of `ConfigOption`, we would need to rewrite/replace things and have the package import `config`. And the `config` package _already imports_ `extpoints`, which would land us a circular reference, and that is not supported by Go.

At the moment, I would leave it to have two identical definitions while we transition to cobra. We'll get rid of `expoints` and its `ConfigOption` in two or three weeks anyway, so it wouldn't be a good use of time to clean that code.

t0xicCode:
Fair enough! I would recommend adding something on the doc line of `extpoints.ConfigOption` in the meantime though.

yannlandry:
Little update. Fixed according to @t0xicCode's comments. Formatted. Disabled soft tabs in my editor. Also renamed `config.Register` to something a bit more unique because I thought it was mean to steal the keyword `Register` for just that.

config/init.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/taskcluster/taskcluster-cli/client"
+)
+
+var (
+	Configuration map[string]map[string]interface{}
+	ConfigOptions = make(map[string]map[string]ConfigOption)
+	Credentials   *client.Credentials
+)
+
+// Setup: call it from main
```
t0xicCode: nit: the first line should be formatted as a complete sentence describing the function.

config/option.go
```diff
@@ -0,0 +1,45 @@
+package config
+
+import (
+	"github.com/taskcluster/taskcluster-cli/extpoints"
+)
+
+// A ConfigOption is something with a default value and a validator.
+// Only requirement is that values are JSON structures.
+type ConfigOption struct {
```
t0xicCode: nit: golint recommends renaming it to something like `Option` because users will be accessing it as `config.ConfigOption`, which stutters.

PR #60, opened 2017-02-02
Replace `tctalker`

t0xicCode:
Release engineering has built a Python app to do some taskcluster chores:
  https://github.com/mozilla/tctalker/blob/master/src/tctalker/tctalker.py
We should support that functionality directly.

* `tctalker status` is covered by #23 
* `tctalker cancel` is as simple as `taskcluster api queue cancelTask $TASKID`
* `tctalker rerun` is  as simple as `taskcluster api queue rerunTask $TASKID`
* `tctalker report_completed` would make a good new command
* `tctalker cancel_graph` too.  Note that it uses the deprecated scheduler API.  A better choice would be to list all tasks in the given taskGroupId and cancel those which are not complete.

### DIFF
--- a/group/actions.go
+++ b/group/actions.go
@@ -1,0 +1,95 @@
+package group
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"sync"
+
+	tcclient "github.com/taskcluster/taskcluster-client-go"
+	"github.com/taskcluster/taskcluster-client-go/queue"
+)
+
+type arguments map[string]interface{}
+
+// SubCommand represents the function interface of the task subcommand.
+type SubCommand func(credentials *tcclient.Credentials, args arguments) bool
+
+func runCancel(credentials *tcclient.Credentials, args arguments) bool {
+	q := queue.New(credentials)
+	groupID := args["<groupId>"].(string)
+
+	tasks := make([]string, 0)
+	continuation := ""
+	for {
+		ts, err := q.ListTaskGroup(groupID, continuation, "")
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "error: could not fetch tasks for group %s: %v", groupID, err)
+			return false
+		}
+
+		for _, t := range ts.Tasks {
+			if t.Status.State == "unscheduled" || t.Status.State == "pending" || t.Status.State == "running" {
+				tasks = append(tasks, t.Status.TaskID)
+			}
+		}
+
+		continuation = ts.ContinuationToken
+		if continuation == "" {
+			break
+		}
+	}
+
+	// Here we use a waitgroup to ensure that we return once all the tasks have
+	// been completed.
+	wg := &sync.WaitGroup{}
+	// The context allows us to exit early if any of the cancellation fails.
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	for _, taskID := range tasks {
+		wg.Add(1)
+		go func(taskID string) {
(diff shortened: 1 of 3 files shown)


group/actions.go
```diff
@@ -0,0 +1,95 @@
...
+			case err := <-c:
+				if err != nil {
+					panic(fmt.Errorf("could not cancel task %s: %v", taskID, err))
+				}
+			}
+			// if we exited normally, we indicate that we completed.
+			wg.Done()
+		}(taskID)
+	}
+	// change the semantics o the waitgroup to close a channel instead
```
AyubMohamed: nit:  spelling mistake on to*

group/actions.go
```diff
@@ -0,0 +1,95 @@
...
+	tcclient "github.com/taskcluster/taskcluster-client-go"
+	"github.com/taskcluster/taskcluster-client-go/queue"
+)
+
+type arguments map[string]interface{}
+
+// SubCommand represents the function interface of the task subcommand.
+type SubCommand func(credentials *tcclient.Credentials, args arguments) bool
+
+func runCancel(credentials *tcclient.Credentials, args arguments) bool {
```
AyubMohamed: Maybe write some high level comments about what's going on with this method. Didn't know about go routines till I reviewed this haha

group/group.go
```diff
@@ -0,0 +1,50 @@
...
+}
+
+func (cmd) Summary() string {
+	return "Group related actions."
+}
+
+func (cmd) Usage() string {
+	return `Group related actions.
+
+Usage:
```
AyubMohamed: The usage block looks off here. Is it supposed to be indented?

group/actions.go
t0xicCode: 👍

group/group.go
t0xicCode: This matches existing behaviour ;)

t0xicCode:
@djmitche That's good. I'm not entirely sure how to properly mock the TaskCluster API endpoints.

escapewindow:
Typo? :)

```
akimoz:/src/gecko/mozilla-inbound/testing/mozharness [13:54:29]  (beetmover-scopes)(beetmover-scopes)
668$ taskcluster group
Usage:
  groupcluster group cancel [--] <groupId>
```

I'm guessing it was a global `s,task,group,g` search/replace :)

t0xicCode:
No, not in this case :)

All tasks have a groupid, although the groupid is the same as the taskid
for most cases (most groups have a single task).

Although we can probably see about supporting cancelling a group based on a
taskid as well.

t0xicCode:
Oh wait, you are correct! I'll fix it

djmitche:
Given that this has seen some real-world use, I think we can merge it -- but we do need more tests!!

t0xicCode:
Weird indeed. Should I mark it as reviewed and merge?

t0xicCode:
Github is being really strange right now

djmitche:
Yeah, if you can mark it as reviewed, go ahead

t0xicCode:
It's not letting me, given that I've authored stuff in there.

I'll see if @yannlandry or someone else on irc can mark it as reviewed.

yannlandry:
I can try to mark it as reviewed; confirm everything is approved before I blindly click Approve?

t0xicCode:
Should be good to go, @djmitche looked it over, and releng started using it
in prod over the weekend.

djmitche:
Thanks!  I wonder why I couldn't approve...

PR #95, opened 2017-02-06
Add fetch and checkout

lteigrob:
Not for review yet.. 

Trying out use of github's metadata  ( {{... }} ) which cannot be tested in Task Creator

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -22,7 +22,11 @@ tasks:
         - '-c'
         - >-
           go get -t github.com/taskcluster/taskcluster-cli/... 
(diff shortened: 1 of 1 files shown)


djmitche:
From the logs:
```
HEAD is now at 7ec9a9d... Add fetch and checkout
```
looks like it worked!

lteigrob:
Cool! Just in case people wanted to see what this would look like done in the Task Creator, you can check out this task: 
https://tools.taskcluster.net/task-inspector/#clo70uIKSIempm0rOfOuAQ/
To run your own this way, just sub out my branch name ("signin_support") and then use the current HEAD for your own branch after "checkout"

PR #96, opened 2017-02-07
Add gometalinter tests

lteigrob:
Trying out some of the other tests we want in .taskcluster.yml. There will likely be more added to this PR shortly.


djmitche:
The TaskCluster task says:
```
# cd /go/src/github.com/taskcluster/taskcluster-cli; git pull --ff-only
```
I have no idea where that could have come from, though.. does gometalinter do a `git pull`??? @walac?

lteigrob:
Yes, seems weird. I have some things I'm going to mess around with tomorrow to troubleshoot. This is my first time seeing this type of error though, and I have run many similar commands in task creator.

lteigrob:
The problem was in trying to checkout and then 'go get -u ./...' in that order. Being on a detached head messed this up. So now we get all our packages first, then checkout, then do the tests. 

This wasn't a problem before because I was testing on branches before PR #67 was merged and didn't need 'go get -u ./...' at all. But after this merge I was getting import errors. So, good problem to have now, as .taskcluster.yml will be more robust for future changes.

lteigrob:
This has some extra test features, using both 'gometalinter' and adding to 'go test':

Gometalinter:
This can check for a variety of Go warning and errors. We've included:
_gotype_ - Syntactic and semantic analysis similar to the Go compiler.
_golint_ - Google's (mostly stylistic) linter.
_deadcode_ - Finds unused code.  -- Just realized I've got this one (ironically) enabled twice.
_staticcheck_ - Statically detect bugs, both obvious and subtle ones.
misspell - Finds commonly misspelled English words.
_go vet --shadow_ - Reports variables that may have been unintentionally shadowed
_gosimple_ - Report simplifications in code.
(Taken from: https://github.com/alecthomas/gometalinter/blob/master/README.md)

Test flags:
_-v_ (verbose): logs and prints all tests
_-race_: checks for race conditions

.taskcluster.yml
```diff
@@ -24,9 +24,15 @@ tasks:
           go get -t github.com/taskcluster/taskcluster-cli/... 
           && cd  /go/src/github.com/taskcluster/taskcluster-cli 
           && git fetch {{ event.head.repo.url }} {{ event.head.ref }} 
-          && git checkout {{ event.head.sha }} 
           && make 
-          && go test ./...
+          && go get -u github.com/alecthomas/gometalinter 
+          && gometalinter --install
+          && go get -u ./...
+          && git checkout {{ event.head.sha }}
```
djmitche: These lines have me a little worried -- it's running `go get -u` on the default branch (the one `go get -t github.com/taskcluster/taskcluster-cli/...` found), then switching to the commit we want to test.  If those have different dependent packages, would this still work?

@walac this is pretty far over my golang head -- can you comment here?

walac: This is to make `gometalinter` happy, as it doesn't look for packages inside vendor. Unfortunately, there is no much more we can do about it, since if we try to do `go get` after `git checkout`, it will fail due to the repo being in a detached state.

djmitche: Does that mean that we're running our tests against the latest versions of everything, rather than the vendored versions?  Or is this just updating gometalinter's own dependencies?

t0xicCode: Would it be possible to use `gometalinter --vendor`?

djmitche:
On the other hand, two golang pros have given this a :+1: so I should probably stop arguing :)

.taskcluster.yml
```diff
@@ -24,9 +24,15 @@ tasks:
...
           && make 
-          && go test ./...
+          && go get -u github.com/alecthomas/gometalinter 
+          && gometalinter --install
+          && go get -u ./...
+          && git checkout {{ event.head.sha }} 
+          && go test -v -race ./...
+          && gometalinter --disable-all --enable=gotype --enable=golint 
+          --enable=deadcode --enable=staticcheck --enable=deadcode 
+          --enable=misspell --enable=vet --enable=vetshadow --enable=gosimple
```
t0xicCode: Also, this doesn't run on any of the sub-packages.

.taskcluster.yml
walac: @djmitche nope, tests run under vendor, just `gometalinter`

@t0xicCode oh, I was not aware of vendor option, thanks. @lteigrob could you please give it a try?

lteigrob: Yep, I'll add in that flag right now.

.taskcluster.yml
```diff
@@ -24,9 +24,15 @@ tasks:
           go get -t github.com/taskcluster/taskcluster-cli/... 
           && cd  /go/src/github.com/taskcluster/taskcluster-cli 
           && git fetch {{ event.head.repo.url }} {{ event.head.ref }} 
-          && git checkout {{ event.head.sha }} 
           && make 
-          && go test ./...
+          && go get -u github.com/alecthomas/gometalinter 
+          && gometalinter --install --force
+          && go get -u ./...
```
t0xicCode: We shouldn't need this anymore.

.taskcluster.yml
lteigrob: Shoot. Right you are.

.taskcluster.yml
```diff
@@ -24,9 +24,15 @@ tasks:
...
+          && git checkout {{ event.head.sha }} 
+          && go test -v -race ./...
+          && gometalinter --vendor --disable-all --enable=gotype --enable=golint
```
t0xicCode: But we might have to run `go install ./...` before this. _gotype_ uses the compiled data that is generated by `install` to run its checks.

.taskcluster.yml
t0xicCode: `go test` if fully compatible with vendoring. It's kinda funny the way it does it. When the code is compiled to be tested, the toolchain will pull the dependencies from _vendor/_, as expected. Then, when `go test` runs the tests, it will actually recurse into the folders in _vendor/_, but because `govendor` removes all of the test files from dependencies, no tests are run.

In essence, it matches the behaviour of un-vendored code.

lteigrob:
Perhaps for discussion in the meeting:

Now that I'm checking subdirectories with gometalinter, my task is failing, I believe due to warnings. You can see this here:
https://tools.taskcluster.net/task-inspector/#d8c67_q0SLCiR61kYcU1Kw/0
(Go to Runs->Run0 and then Logs, and scrolling down you can see warnings with the type given in parentheses on the right of each)
where this task is the same but I've just omitted the specific gometalinter tests giving the warnings, and it's fine:
https://tools.taskcluster.net/task-inspector/#WH-069lNSCqkXdvOJibOCQ/0

From what I can tell this is due to gometalinter setting the exit status to 1 for any such warning. Is this what we want? If so, we'd just have to go through and fix up all these warnings to get everything up to this level of strictness. Alternatively, we allow these warnings but figure out how to keep exit status as zero. I'm not sure there's a way to do this using gometalinter as-is.

I'm sure it will fail, but I'll push the change for discussion purposes :) I'd like to talk more about @t0xicCode's comments too

t0xicCode:
I'm actually building a PR to fix most of the linting issues.

lteigrob:
Oh, perfect!

lteigrob:
Nice, I'll just figure out the go install vs. go get -u issue and we should be good

t0xicCode:
@lteigrob Could you rebase on top of master? I've merged the linting fixes that you approved :)

lteigrob:
_One does not simply git rebase it_... @t0xicCode,  I also had to fetch and merge with the remote gometalinter branch. Not sure if that's normal. I've looked at the code and I don't think I killed anything.

t0xicCode:
I may have managed to do it 😄 but I manually removed a bunch of duplicated commits.

.taskcluster.yml
```diff
@@ -24,9 +24,15 @@ tasks:
           go get -t github.com/taskcluster/taskcluster-cli/... 
           && cd  /go/src/github.com/taskcluster/taskcluster-cli 
           && git fetch {{ event.head.repo.url }} {{ event.head.ref }} 
-          && git checkout {{ event.head.sha }} 
           && make 
-          && go test ./...
+          && go get -u github.com/alecthomas/gometalinter 
+          && gometalinter --install --force
+          && go install ./...
+          && git checkout {{ event.head.sha }}
```
t0xicCode: You should checkout before running `go install` or `make`, in case the dependencies changed.

t0xicCode:
Also, you'll have to do a force pull (`git fetch && git reset --hard origin/gometalinter`) before continuing.

.taskcluster.yml
```diff
@@ -24,9 +24,15 @@ tasks:
...
           && git fetch {{ event.head.repo.url }} {{ event.head.ref }} 
-          && git checkout {{ event.head.sha }} 
           && make 
-          && go test ./...
+          && go get -u github.com/alecthomas/gometalinter 
+          && gometalinter --install --force
+          && go install ./...
+          && git checkout {{ event.head.sha }} 
+          && go test -v -race ./...
+          && gometalinter --vendor --disable-all --enable=gotype --enable=golint
```
djmitche: Consider --skip=apis/ to skip the auto-generated source (for #100)

lteigrob:
Cool, this sequence seems to be working now. I'm sure @djmitche will be happy :)

Aaand I got excited and forgot to do the force pull :( If things need to be fixed we can discuss tomorrow. 

Finally, I'm working on getting '--skip=apis' to work. Like everything else since adding in gometalinter, it's not as simple as it seems.

lteigrob:
Cool, this sequence seems to be working now. I'm sure @djmitche will be happy :)

I've played around with getting '--skip=apis' to work. Not as easy as just adding in the flag, but I'll get it figured.

djmitche:
Success on Taskcluster!!  Should we merge this?

djmitche:
I can just squash it when merging.. that'll be easier, I think.

t0xicCode:
Yeah, that works too

PR #97, opened 2017-02-09
Tailrunningtask#51

AyubMohamed:
As of right now this command will only tail a task for a task that has completed. There is an issue with tailing live tasks because the `public/logs/live_backing.log` file hasn't been created yet. Maybe we can figure out a way to connect to the worker and get live logs that way but I have been unable to do it so far.


task-log/task_log.go
```diff
@@ -0,0 +1,77 @@
...
+	route := apis.Services["Queue"].BaseURL
+	for _, entry := range apis.Services["Queue"].Entries {
+		if entry.Name == "getLatestArtifact" {
+			route += entry.Route
+			break
+		}
+	}
+
+	route = strings.Replace(route, "<taskId>", taskID, 1)
+	route = strings.Replace(route, "<name>", "public/logs/live_backing.log", 1)
```
djmitche: I think this should use `live.log` instead.  For completed tasks, that seems to work fine in my testing: the `live.log` artifact redirects to `live_backing.log`.

task-log/task_log.go
```diff
@@ -0,0 +1,77 @@
...
+
+	body, _ := makeGetRequest(route)
+
+	// Check if we got any errors, we are not expecting a json response.
+	var raw map[string]interface{}
+	json.Unmarshal(body, &raw)
+
+	if len(raw) != 0 {
+		// Error, most likely with the taskID
```
djmitche: Good thinking!

task-log/task_log.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+func makeGetRequest(path string) (b []byte, err error) {
+	resp, err := http.Get(path)
+	if err != nil {
+		panic("Error making request to " + path)
+	}
+
+	defer resp.Body.Close()
+	body, err := ioutil.ReadAll(resp.Body)
```
djmitche: To tail a log "live", I think you'll only need to change this from `ReadAll` to a loop that reads whatever is available form resp.Body, then writes that data to stdout.  That will be a change to the `makeGetRequest` function, since it can no longer return a string.  But that makes sense: if we want to print parts of the task log before the rest of the log exists, we can't return the whole log!

AyubMohamed: The problem is that when we try to tail a "live" log, the get request hangs until the task is completed.

task-log/task_log.go
AyubMohamed: For running tasks using `live.log`,  the get request will hang until the task is completed. I chose to use `live_backing.log` for this pr because we don't have a proper way to deal with live tasks at the moment. So in this case using `live_backing.log` will at least give us an error for running tasks.

jonasfj: Yes, `live.log` is a streaming HTTP request, just read it as a stream...
If you're fancy you can do some output buffering before writing to stdout.

task-log/task_log.go
```diff
@@ -0,0 +1,77 @@
...
+func (taskLog) Usage() string {
+	usage := "Usage: taskcluster task-log <taskID>\n"
+	return usage
+}
+
+func (taskLog) Execute(context extpoints.Context) bool {
+	taskID := context.Arguments["<taskID>"].(string)
+
+	// Get route from services.go
+	route := apis.Services["Queue"].BaseURL
```
jonasfj: Yeah, I would not do this...

Client for golang has a signed url method: https://godoc.org/github.com/taskcluster/taskcluster-client-go/queue#Queue.GetLatestArtifact_SignedURL
But since we don't need a signed url here.. I would just hardcode the URL pattern...

jonasfj:
@t0xicCode The tcclient doesn't support streaming responses, and interprets responses as JSON... It's rather domain specific that way.

But yeah, the requests should be done retries with exponential back-off.. I would just look at how that's done in `go-got` (disclaimer I wrote that).

t0xicCode:
@jonasfj I was thinking: tcclient to generate the URL, then using go-got or the standard http client to fetch the data and print it (maybe buffering it). 

Does that sound like a feasible thing? Or is it "overkill"?

jonasfj:
Is there a method to generate URL... I could only find the one to generate
a signed URL...

Den 14. feb. 2017 6.17 PM skrev "Xavier L." <notifications@github.com>:

t0xicCode:
Seems like there isn't. Is it worth it adding such a function to tcclient?

jonasfj:
Probably, maybe @petemoore knows a reason we never added it...

Den 14. feb. 2017 6.24 PM skrev "Xavier L." <notifications@github.com>:

djmitche:
Why not use a signed URL?  It doesn't hurt, and it's something the library already supplies for us..

@AyubMohamed: in the middle of all this back-and-forth, I think the immediate next step is to get streaming working using the URL you've already written the code for.  If we figure out a better way to invent a URL, that should be an easy switch later.

task-log/task_log.go
AyubMohamed: That's what I did initially, but i thought we should use that services struct in case the api changes or anything. But i'll just hardcode it for now.

apis/doc.go
```diff
@@ -10,7 +10,7 @@ import (
 )
 
 func init() {
-	for name, service := range services {
+	for name, service := range Services {
```
djmitche: I think you can revert this change and the change to `services.go` now that you're not using that data in `task_log.go`, right?

task-log/task_log.go
```diff
@@ -0,0 +1,71 @@
...
+	defer resp.Body.Close()
+
+	// Read line by line for live logs.
+	// This will also print the error message for failed requests.
+	scanner := bufio.NewScanner(resp.Body)
+	for scanner.Scan() {
+		fmt.Println(scanner.Text())
+	}
+
+	body, err := ioutil.ReadAll(resp.Body)
```
djmitche: I'm assuming this always returns an empty string, since you've already read everything in the for loop, above.  So maybe this function should just return an error.  Which means the unmarshalling in Execute isn't necessary.

However, thinking about the error handling here -- I think that Execute would return `true` even for a taskId that doesn't exist, right?  Probably `makeGetRequest` should be looking at the status code in `resp` and signalling an error when the code is not 200 (OK).  In the error case, writing the JSON body to stderr is probably a good approach.

AyubMohamed:
Live logging support now works, I think once we figure out a good way to construct the URL we can create another PR for that issue.

petemoore:
Strange, I thought I added a comment here, but I can't see it now! Anyway, I also think using SignedURL should be fine, and do no harm.

jonasfj:
> Why not use a signed URL? It doesn't hurt, and it's something the library already supplies for us..

Because then you would have to authenticate inorder to view a public log... that's seems excessive.
Many people will likely authenticate with tc-cli, but we shouldn't make unauthenticated APIs require login when it's not needed...

task-log/task_log.go
```diff
@@ -0,0 +1,50 @@
...
+	if err != nil {
+		panic("Error making request to " + path)
+	}
+
+	defer resp.Body.Close()
+
+	// Read line by line for live logs.
+	// This will also print the error message for failed requests.
+	scanner := bufio.NewScanner(resp.Body)
+	for scanner.Scan() {
```
jonasfj: Is there any reason to read it line by line?

What if I have a task that plays a cat video as ASCII art using VLC...
Hehe :)

I don't think terminal progress bars like the ones hg or git displays when cloning are necessarily line separated.
Why not just do: `io.Copy(os.Stdout, resp.Body)` ?
If smart you would track length too, match it against content-length if given, and use the length to cut away bytes if retrying the request... So logs don't get duplicated on retries. 


If you want to be really fancy you'll copy from `resp.Body` to an intermediate buffer and flush from that buffer to `os.Stdout` at an interval... Example here:
https://github.com/taskcluster/taskcluster-worker/blob/master/runtime/ioext/copyflush.go
If the intermediate buffer implemented Flush() this CopyAndFlush method could be used.

Oh, and you might want to look into setting stdout into raw mode... Not sure all our logs are actually in raw mode, so maybe we would have to detect that... I think some of the logs uses `\r\n` as line separator which is a good indicator of raw mode.

petemoore:
@jonasfj We should be mindful that one day logs might also be private (and thus signed urls make sense). Maybe we should update the methods that return signed urls to only return signed versions if scopes are actually required, and raw (unsigned) urls if no scopes are required to avoid an additional redirect.

djmitche:
Regarding "fancy", I think line-by-line is good for logging, and this is
indeed a log.  I don't think we need to retry these requests.

jonasfj:
@petemoore 
> Maybe we should update the methods that return signed urls to only return signed versions if scopes are actually required, and raw (unsigned) urls if no scopes are required to avoid an additional redirect.

This can't be determined in the client.

----
To support secret logs, we could add support for options like `--log-path`, that's only specified if it's not `public/live.log`.

Similarly, it would be nice to have an optional `--run-id <runId>` option in case you don't want the latest runId...

djmitche:
OK, so a few follow-on issues to open:
 * generate the URL differently (still an open question on how, or whether, to do this)
 * `--log-path` and `--run-id`
I've added the first point to the agenda for Friday's meeting.

task-log/task_log.go
AyubMohamed: The problem with io.Copy is that it copies from src to dst until EOF is reached. For live logging the request doesn't come back until the log is complete, so there is no EOF at the moment.

djmitche: I think io.Copy would probably work, but it might return chunks of log that aren't newline-aligned, which is sort of ugly.

I think for logs, which are line-oriented, the line-by-line approach is right.

AyubMohamed: Actually my mistake, this should work. I just don't know how big the chunk or how often it copies to `os.Stdout`.

PR #98, opened 2017-02-10
Fix some linting issues

t0xicCode:
This fixes a lot of the linting issues that @lteigrob found with gometalinter.


t0xicCode:
FYI, this will need to be rebased if accepted after the cobra changes.

PR #103, opened 2017-02-18
Port config command to cobra

yannlandry:
I've created a subbranch of **install-cobra** to port `taskcluster config` and its subcommands.

I've split the `config` package into two packages: one contains the logic required to manage the configuration (save, load, (un)serialize, register options and keep configuration), and the other is for `taskcluster config` and its subcommands.


t0xicCode:
quick look seems fine. I'll do a more in depth review in a bit! 👍

config-cmd/config.go
```diff
@@ -0,0 +1,110 @@
...
+	"os"
+
+	"github.com/spf13/cobra"
+	"github.com/taskcluster/taskcluster-cli/config"
+	"github.com/taskcluster/taskcluster-cli/root"
+	"github.com/taskcluster/taskcluster-client-go"
+)
+
+var (
+	Command = &cobra.Command{
```
t0xicCode: nit: exported var Command should have comment or be unexported (golint)

config-cmd/help.go
```diff
@@ -0,0 +1,98 @@
...
+func printHelp(cmd *cobra.Command) {
+	commands := []string{}
+	maxLength := 0
+
+	for command, options := range config.OptionsDefinitions {
+		// this is for the list of commands
+		commands = append(commands, command)
+
+		// this is for the maximum name length for a key
+		for option, _ := range options {
```
t0xicCode: nit: should omit 2nd value from range; this loop is equivalent to `for option := range ...` (golint)

config-cmd/signin.go
```diff
@@ -0,0 +1,127 @@
...
+	fmt.Fprintln(cmd.OutOrStdout(), "Listening for a callback on: "+target)
+	webbrowser.Open(loginURL, webbrowser.NewWindow, true)
+
+	// Start serving
+	s.Serve(listener)
+
+	if serr != nil {
+		return fmt.Errorf("failed to save configuration, error: %s\n", serr)
+	} else {
+		return nil
```
t0xicCode: nit: if block ends with a return statement, so drop this else and outdent its block (golint)

config-cmd/helpers.go
```diff
@@ -0,0 +1,84 @@
...
+	return command, option, definition, value, err
+}
+
+// parseKey parses a key into command, option
+// use if you don't need to retrieve the definition
+func parseKey(key string) (string, string, error) {
+	parts := strings.SplitN(key, ".", 2)
+
+	if len(parts) != 2 || len(parts[0]) == 0 || len(parts[1]) == 0 {
+		return "", "", fmt.Errorf("invalid key format '%s', configuration keys must be on the form '<command>.<option>'")
```
t0xicCode: nit: missing argument for Errorf("%s"): format reads arg 1, have only 0 args (vet)

config-cmd/help.go
```diff
@@ -0,0 +1,98 @@
...
+func cmdHelp(cmd *cobra.Command, args []string) error {
+	// default, print help for all commands
+	if len(args) == 0 {
+		printHelp(cmd)
+		return nil
+	}
+
+	// otherwise, print for one specific option
+	err := printOptionHelp(cmd, args[0])
+	if err != nil {
```
t0xicCode: nit: 'if err != nil { return err }; return nil' can be simplified to 'return err' (gosimple)

config-cmd/signin.go
```diff
@@ -0,0 +1,127 @@
+package configCmd
```
t0xicCode: Would it make more sense to have the signin command in a different package? As a new contributor, I probably wouldn't expect it to be defined with the config commands.

config/option.go
```diff
@@ -14,8 +14,8 @@ type OptionDefinition struct {
 	Validate    func(value interface{}) error
 }
 
-// RegisterConfigOption takes in the name of the command and an OptionDefinition object
-func RegisterConfigOption(command string, options map[string]OptionDefinition) {
```
t0xicCode: Make sure that you've renamed it throughout the application, seems like travis found a place were it wasn't :)

subtree_imports.go
```diff
@@ -5,6 +5,7 @@ package main
 
 import _ "github.com/taskcluster/taskcluster-cli/apis"
 import _ "github.com/taskcluster/taskcluster-cli/config"
+import _ "github.com/taskcluster/taskcluster-cli/config-cmd" // commands for config
```
t0xicCode: I still would advocate to move all the commands into a `cmds` subpackage, so the config commands would be in the `github.com/taskcluster/taskcluster-cli/cmds/config` package. I believe that it would make the layout and separation cleaner.

t0xicCode:
@yannlandry, think you can rebase on top of the latest `install-cobra`?

subtree_imports.go
yannlandry: It's starting to make sense. I'm gonna keep the current package as it is for now, but we'll talk about this again before merging `install-cobra` into `master`.

config-cmd/signin.go
yannlandry: I'll move it. Wasn't sure it was useful to create a whole package for a single file/command.

yannlandry:
@t0xicCode missed your previous comment, let's see what I can do

t0xicCode:
Perfect

Issue #49, opened 2017-01-11
Build releases using taskcluster-github


djmitche:
When we create a new Github release of taskcluster-cli, tc-github should automatically build and upload binaries for the various platforms we support, and make them available for users.

djmitche:
I can get this working for linux releases, and then hand off to @t0xicCode to set up other platforms.  Hopefully we can cross-compile for those platforms!

djmitche:
There's no way to upload those artifacts to a github release, which is too bad because that'd be the most convenient!
  https://bugzilla.mozilla.org/show_bug.cgi?id=1344913

djmitche:
My current plan is this:
 * react to "release" events and build the corresponding tag
 * on release, upload the resulting binary to an S3 bucket with an appropriately versioned name, and also update a "latest" value
 * point to the "latest" URL from the README

After #107, any cross-compilations should be easy to include in this process, if you want to get started on that @t0xicCode

djmitche:
(The part remaining for me is points 2 and 3)

djmitche:
Working on the second point, I got stuck at #122.

djmitche:
OK, that's done -- all that's left is the cross-compilation!

t0xicCode:
Perfect! I guess I'll take over from here :)

PR #107, opened 2017-03-06
add a build that uploads artifacts

djmitche:
This task, when complete, should have a `taskcluster-linux` artifact containing an executable binary.  The idea is that we can add a darwin (OS X) and windows binary, too, using similar tasks.

When this is set, I'll change it to run on releases, rather than on pull requests (which should still just run tests).

### DIFF
--- a/cmds/version/version.go
+++ b/cmds/version/version.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 
 	"github.com/spf13/cobra"
-
 	"github.com/taskcluster/taskcluster-cli/cmds/root"
 )
 
@@ -16,8 +15,9 @@ var (
 		Run:   printVersion,
 	}
 
-	// VersionNumber is a formatted string with the version information.
-	VersionNumber = "1.0.0"
+	// VersionNumber is a formatted string with the version information. This is
+	// filled in by the Makefile
+	VersionNumber string
 )
 
 func init() {
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -1,7 +1,7 @@
 version: 0
 metadata:
-  name: "TaskCluster GitHub Tests"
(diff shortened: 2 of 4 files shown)


djmitche:
Indeed, it seems to have built a binary: https://tools.taskcluster.net/task-inspector/#YBsqfe-aRVucm2ftlO6hBA/

djmitche:
@t0xicCode let me know what you think; note that this doesn't complete #49 but it's a start.

cmds/version/version.go
```diff
@@ -16,8 +15,9 @@ var (
 		Run:   printVersion,
 	}
 
-	// VersionNumber is a formatted string with the version information.
-	VersionNumber = "1.0.0"
+	// VersionNumber is a formatted string with the version information. This is
+	// filled in by the Makefile
+	VersionNumber string
```
t0xicCode: The matching test should probably be updated here.

t0xicCode:
@djmitche have you looked at https://github.com/mitchellh/gox ? It's basically a wrapper around go's build that takes care of the right arguments for building.

djmitche:
> @djmitche Did we say that for the darwin and windows binaries we'd have to change some other sections of .taskcluster.yml as well?

I think we can build darwin and windows binaries on linux -- I hope so, anyway, since we don't have any mac systems we can run these builds on!

> @djmitche have you looked at https://github.com/mitchellh/gox ? It's basically a wrapper around go's build that takes care of the right arguments for building.

No, but that looks cool -- I'm leaving the cross-compiling to you :)

cmds/version/version.go
djmitche: It looks for `VersionNumber` in the returned string, so it will always find it.

PR #109, opened 2017-03-06
Move root command to cmds/ subfolder

yannlandry:
It seemed like a logical thing to do, because like, it's a command.


cmds/group/actors.go
```diff
@@ -1,14 +1,14 @@
...
 import (
-	"context"
 	"fmt"
 	"io"
 	"sync"
 
 	"github.com/spf13/pflag"
 	tcclient "github.com/taskcluster/taskcluster-client-go"
 	"github.com/taskcluster/taskcluster-client-go/queue"
+	"golang.org/x/net/context"
```
djmitche: This change looks odd..

yannlandry: Yes, we used to simply import `"context"` instead of that whole address. I changed it to `"golang.org/x/net/context"`, as it's the same package, but compatible with older versions of Go.

I've scheduled for us to discuss which version of Go should be required to compile the project at our next meeting.

PR #110, opened 2017-03-07
Use backwards compatible version of package 'context'

yannlandry:
This is an optional change that would enable us to compile with older versions of Go.

_Details:_

Currently, the file **cmds/group/actors.go** imports package `context`. However, that package is only available "natively" in Go starting at version 1.7 (I believe).

Older versions of Go can access the `context` package by importing `golang.org/x/net/context`. Everything else will work the same.

This change makes tc-cli use the latter form (long URL) so older versions of Go can compile the program. Of which version 1.5.1, the current version available in Ubuntu's (and perhaps others') default repositories. Users that want a later version of Go must install it manually. It's up to us to decide whether we want to force users to install Go > 1.7 - potentially manually - or keep backwards compatibility. I've also added that point to Friday's meeting.

### DIFF
--- a/cmds/group/actors.go
+++ b/cmds/group/actors.go
@@ -1,7 +1,6 @@
 package group
 
 import (
-	"context"
 	"fmt"
 	"io"
 	"sync"
@@ -9,6 +8,7 @@ import (
 	"github.com/spf13/pflag"
 	tcclient "github.com/taskcluster/taskcluster-client-go"
 	"github.com/taskcluster/taskcluster-client-go/queue"
+	"golang.org/x/net/context"
 )
 
 // runCancel cancels all tasks of a group.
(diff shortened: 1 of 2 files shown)


walac:
It works on travis but failed on taskcluster, any idea why?

djmitche:
I imagine "context" needs to be added to vendor.json? 

I don't mind restricting to Go-1.7+, but if allowing 1.5.1 is as easy as this, that's good too.  What do others think?

walac:
> I don't mind restricting to Go-1.7+, but if allowing 1.5.1 is as easy as this, that's good too. What do others think?

I have no objections.

yannlandry:
@djmitche yeah I forgot about the vendoring thing, should be good now.

PR #112, opened 2017-03-08
Use standard library implementation for fetching user home directory

yasch007:
@jonasfj There looks to be a simple way to determine a user's home directory in the standard library, which simplifies the implementation a little.

### DIFF
--- a/config/serialization.go
+++ b/config/serialization.go
@@ -4,27 +4,22 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"log"
 	"os"
+	"os/user"
 	"path/filepath"
 	"reflect"
 
-	homedir "github.com/mitchellh/go-homedir"
 	yaml "gopkg.in/yaml.v2"
 )
 
 // configFile is the location of the configuration file
 func configFile() string {
-	configFolder := os.Getenv("XDG_CONFIG_HOME")
-	if configFolder == "" {
-		homeFolder := os.Getenv("HOME")
-		if homeFolder == "" {
-			homeFolder, _ = homedir.Dir()
-		}
-		if homeFolder != "" {
-			configFolder = filepath.Join(homeFolder, ".config")
-		}
+	u, err := user.Current()
+	if err != nil {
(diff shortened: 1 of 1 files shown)


config/serialization.go
```diff
@@ -4,27 +4,22 @@ import (
...
-		if homeFolder != "" {
-			configFolder = filepath.Join(homeFolder, ".config")
-		}
+	u, err := user.Current()
+	if err != nil {
+		log.Fatalf("Cannot determine user home directory required for storing taskcluster-cli configuration file: %v", err)
 	}
-	return filepath.Join(configFolder, "taskcluster.yml")
+	return filepath.Join(u.HomeDir, "taskcluster.yml")
 }
```
jonasfj: I doubt u.homeDir is the the config directory from xdg-folder specification... Which we should follow on Linux at least...

t0xicCode: As well, from https://github.com/mitchellh/go-homedir: 

> **Why not just use `os/user`?** The built-in `os/user` package requires
cgo on Darwin systems. This means that any Go code that uses that package
cannot cross compile. But 99% of the time the use for `os/user` is just to
retrieve the home directory, which we can do for the current user without
cgo. This library does that, enabling cross-compilation.

djmitche:
So, to summarize:
 * the original code determines a path to a configuration file in a subdirectory of the user's home directory, but the patch places that file in the home directory.  The function is already returning the correct path according to [the xdg standard](https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html), so that shouldn't change.
 * `os/user` requires cgo on darwin, which would mean we can't cross-compile, so we would rather not use it.
 * There are is some discussion of whether the XDG standard makes sense for a CLI application.

I can say that I use a number of CLI tools (bugwarrior among them) that do follow this convention, and I think it's reasonable.

So, I'm not sure there's anything to fix here :(

yasch007:
Thanks guys and sorry for the noise.

t0xicCode:
Thanks for your PR! Don't hesitate to submit other PRs/comment on existing
issues/PRs if you see something that you can help with or fix.

We are happy to get all the help we can :)

Issue #111, opened 2017-03-07
Test the API subcommands generation and execution


yannlandry:
There was no specific issue for this. I'm currently on it.

See branch [`test-apis`](https://github.com/taskcluster/taskcluster-cli/tree/test-apis) for details and current progress.

PR #113, opened 2017-03-08
Test the API subcommands generation and execution

yannlandry:
Addressing my own issue #111.

First, I refactored the code-generation code and the command-generation code so it could be tested.

Then, wrote 2 tests.

**TestCodeGeneration**

This test creates a server (in a goroutine) that can serve a manifest and an api definition. It then launches the code-generation function, which in turn calls the test server and generates code. We match the generated code against the contents of **services_test.go**.

It currently does not do schemas generation and tests a very simple definition.

**TestCommandGeneration**

This test also creates a server that can match a request against something predefined, and reply to tell whether the request was correct or not.

Currently, the request being matched is a very simple GET request, and we simply test whether or not a request was received.


yannlandry:
Build is failing:

```
apis/codegen_test.go:28: server.Close undefined (type *http.Server has no field or method Close)
apis/provider_test.go:49: server.Close undefined (type *http.Server has no field or method Close)
```

Anyone has any idea why this is happening?

djmitche:
The missing Close is definitely weird -- it *does* have such a method, at least [according to godocs](https://golang.org/pkg/net/http/#Server.Close).  I don't know how to look at godocs for old versions of go, but perhaps in go-1.7.3 (what's running on Travis) this doesn't exist, while in go-1.8 it does?  @petemoore any help?

yannlandry:
It seems that Close [didn't exist before 1.8](https://beta.golang.org/doc/go1.8#http_shutdown)... the more you know.

Guess we'll have to force Go 1.8 for compilation _or_ I find a way to shutdown the server that's compatible with older versions (preferrable).

apis/codegen_test.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+// manifestServer sets up the server before launching it in a new thread
+func manifestServer() *http.Server {
+	handler := http.NewServeMux()
+	handler.HandleFunc("/manifest.json", manifestHandler)
+	handler.HandleFunc("/definition.json", apiDefHandler)
+
+	server := &http.Server{
+		Addr:    ":8080",
```
yannlandry: The `:8080` address is hardcoded into **services_test.go** and **codegen_test.go**; I feel it would be a pain to have to fix everything just like that :/

yannlandry: Wait, looking back at it, I think it's actually possible to make it work. I'm on it.

yannlandry:
Redid the test servers using `httptest`. That thing is magic. We can ensure that we use an open port for the test server, _and_ the build passes on probably most versions of Go.

djmitche:
🐰  🎩 !

yannlandry:
The content is in that file as Go code because the code has to be reused as part of `TestCommandGeneration`.

djmitche:
Could you include it in provider_test.go?  Maybe between `/** START GENERATED **/` and `/** END GENERATED **/` or something like that?  I like the idea that one test verifies json -> go, and the other verifies go -> functionality.

t0xicCode:
I would argue to make it a string literal in the other test file because
it's data, test data.

djmitche:
Yes, but that same test data is executed as part of another test.  If it's
a string literal, then we'd need to have *another* test case to compare the
string literal to the code executed in that test.

yannlandry:
Not necessarily. The data used for the second test (go -> functionality) can be different from the data generated by the first test (json -> go). In fact, it might be preferable, because we can probably run the first test with only one test case, while the second might require multiple test cases.

I'll put the data to be matched after test 1 in a string literal, and use different data for test 2, that I'll declare in **provider_test.go**.

apis/codegen_test.go
```diff
@@ -30,9 +29,7 @@ func TestCodeGeneration(t *testing.T) {
 
 	// check that the returned byte thing is correct
 	generated := strings.Trim(string(source), "\n\r\t ")
-	desired, err := getDesiredOutput("services_test.go")
-
-	assert.NoError(err, fmt.Sprintf("error getting desired output: %s", err))
+	desired := getDesiredOutput("services_test.go")
 	assert.Equal(generated, desired, "generated code doesn't match desired code")
```
t0xicCode: Here, the format should be `assert.Equal(expected, actual, msg)` :) https://godoc.org/github.com/stretchr/testify/require#Assertions.Equal

apis/codegen_test.go
```diff
@@ -78,16 +75,36 @@ func apiDefHandler(w http.ResponseWriter, _ *http.Request) {
...
-	if err != nil {
-		return "", fmt.Errorf("unable to read %s, error: %s", filename, err)
-	}
-
-	match := strings.Split(string(contents), "//###START###\n")[1]
-	match = strings.Trim(match, "\n\r\t ")
-
-	return match, nil
+// getDesiredOutput simply returns the code that should be generated
+func getDesiredOutput(filename string) (string) {
```
t0xicCode: Could we drop the parameter? (and the parentheses around `string`?)

djmitche: does this still need a filename?

PR #115, opened 2017-03-09
Test for task log

AyubMohamed:


cmds/task/fetchers_test.go
```diff
@@ -0,0 +1,35 @@
...
+	"github.com/spf13/cobra"
+	assert "github.com/stretchr/testify/require"
+)
+
+func TestLogCommand(t *testing.T) {
+	assert := assert.New(t)
+
+	buf := &bytes.Buffer{}
+	cmd := &cobra.Command{}
+	cmd.SetOutput(buf)
```
djmitche: This is great :)

djmitche:
To be clear where this fits in the general testing of taskcluster-cli: I *like* that this tests against a real-life task, since it does its own HTTP handling (to support streaming).  This is probably one of the only subcommands that will need this kind of testing -- for everything else, I think we can use a mock http server as we're discussing on other issues and in email.

Issue #81, opened 2017-01-24
Replace javascript taskcluster-cli (Allow task creation, scheduling, tailing)


t0xicCode:
A previous attempt at a command line client in javascript exists (see [gregarndt/taskcluster-cli](https://github.com/gregarndt/taskcluster-cli)).

This client should be able to replace that javascript implementation. 

* `taskcluster task create` to replace `taskcluster run`: creates and schedules a task
* `taskcluster task run` to replace `taskcluster run-task`: creates, schedules, and watches a task (streaming logs from _public/logs/live.log_ by default)
* `taskcluster task await` to wait the completion of a given task (by taskId)

This depends on #23 and is related to #60.

djmitche:
Good find!  Yes, the first two represent good functionality.

`run-graph` refers to the [task-graph scheduler](https://github.com/taskcluster/task-graph-scheduler), which is deprecated -- instead, graphs are now constructed task-by-task, with the dependencies included in the task definition.  So I don't think there's much need to reproduce that functionality.

t0xicCode:
Sounds good. I'll remove it.

PR #118, opened 2017-03-10
Add task creation

t0xicCode:
This is the first in a series of smaller PRs addressing #81. See https://github.com/gregarndt/taskcluster-cli/blob/master/bin/taskcluster-run.js for the initial implementation.


cmds/task/create.go
```diff
@@ -0,0 +1,138 @@
...
+	"github.com/taskcluster/taskcluster-cli/config"
+	tcclient "github.com/taskcluster/taskcluster-client-go"
+	"github.com/taskcluster/taskcluster-client-go/queue"
+)
+
+var (
+	createCmd = &cobra.Command{
+		Use:     "create <image> <command>",
+		Short:   "Creates and schedules a task, using the specified taskId if given.",
+		PreRunE: checkCreateFlags,
```
yannlandry: Any reason we're using the PreRun, or that's only for code readability?

cmds/task/create.go
```diff
@@ -0,0 +1,138 @@
...
+	tags["misc_info"] = "created by taskcluster-cli"
+	createPayload.Tags, err = json.Marshal(tags)
+	if err != nil {
+		return fmt.Errorf("could not marshal tags: %v", err)
+	}
+
+	q := queue.New(creds)
+	resp, err := q.CreateTask(taskID, createPayload)
+	if err != nil {
+		return fmt.Errorf("Could not create task: %v", err)
```
yannlandry: nit: I thought error messages shouldn't start with a capital letter ;-)

cmds/task/utils.go
```diff
@@ -35,3 +35,11 @@ func executeHelperE(f Executor) func(*cobra.Command, []string) error {
 		return f(creds, args, cmd.OutOrStdout(), cmd.Flags())
 	}
 }
+
+func stringFlagHelper(flagset *pflag.FlagSet, flag string) string {
+	val, err := flagset.GetString(flag)
+	if err != nil {
+		panic(fmt.Errorf("error: error encountered getting the value of %s: %v", flag, err))
```
yannlandry: nit: if we chain this error, we're gonna end with way too many times the word 'error' in a row!

cmds/task/create.go
t0xicCode: It's mostly to separate the checking of the flags from the actual execution of the command.

That way, the command can run assuming that the flag was defined and passed in the command line.

cmds/task/create.go
t0xicCode: you are correct :)

cmds/task/create.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+
+	now = time.Now().UTC()
+
+	createPayload = &queue.TaskDefinitionRequest{
+		SchedulerID: "taskcluster-cli",
+
+		Created:  tcclient.Time(now),
+		Deadline: tcclient.Time(now.Add(24 * time.Hour)),
+		Expires:  tcclient.Time(now.Add(24*time.Hour).AddDate(1, 0, 0)),
```
djmitche: Of these, I can see a user wanting to specify schedulerId, deadline, and expires (the latter two preferably with a from_now format).

That said, I get the impression this is implementing the way this works in the JS taskcluster-cli.

cmds/task/create.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+)
+
+func init() {
+	fs := createCmd.Flags()
+
+	fs.String("task-id", "", "taskID of the task, if omitted, a new one will be assigned")
```
djmitche: IMHO this should always be assigned automatically.

cmds/task/create.go
```diff
@@ -0,0 +1,139 @@
...
+	fs.StringVar(&createPayload.ProvisionerID, "provisioner", "", "ID of the provisioner to use")
+	fs.StringVar(&createPayload.WorkerType, "worker-type", "", "worker-type to use within the provisioner")
+	fs.StringSliceP("env", "e", []string{}, "Environment variable to add to the task's environment (repeatable) (format: VARIABLE=VALUE)")
+	fs.StringVar(&createPayload.Metadata.Name, "name", "", "Human readable name of the task")
+	fs.StringVar(&createPayload.Metadata.Description, "description", "", "Human readable description of the task")
+	fs.StringVar(&createPayload.Metadata.Owner, "owner", "", "Email of the task's owner")
+	fs.StringVar(&createPayload.Metadata.Source, "source", "", "URL pointing to the source of the task")
```
djmitche: These four should have a default.  Note that owner needs to look like an email.

cmds/task/create.go
```diff
@@ -0,0 +1,139 @@
...
+	fs.StringSliceVar(&createPayload.Dependencies, "dependency", []string{}, "TaskID of a dependency (repeatable)")
+	for _, f := range []string{"provisioner", "worker-type"} {
+		createCmd.MarkFlagRequired(f)
+	}
```
djmitche: This probably isn't necessary

cmds/task/create.go
```diff
@@ -0,0 +1,139 @@
...
+	fs.String("task-id", "", "taskID of the task, if omitted, a new one will be assigned")
+	fs.StringVar(&createPayload.ProvisionerID, "provisioner", "", "ID of the provisioner to use")
+	fs.StringVar(&createPayload.WorkerType, "worker-type", "", "worker-type to use within the provisioner")
+	fs.StringSliceP("env", "e", []string{}, "Environment variable to add to the task's environment (repeatable) (format: VARIABLE=VALUE)")
+	fs.StringVar(&createPayload.Metadata.Name, "name", "", "Human readable name of the task")
+	fs.StringVar(&createPayload.Metadata.Description, "description", "", "Human readable description of the task")
+	fs.StringVar(&createPayload.Metadata.Owner, "owner", "", "Email of the task's owner")
+	fs.StringVar(&createPayload.Metadata.Source, "source", "", "URL pointing to the source of the task")
+	fs.StringSliceVar(&createPayload.Dependencies, "dependency", []string{}, "TaskID of a dependency (repeatable)")
+	for _, f := range []string{"provisioner", "worker-type"} {
```
djmitche: This list appears in two places -- constant?

cmds/task/create.go
```diff
@@ -0,0 +1,139 @@
...
+	if err != nil {
+		return fmt.Errorf("could not marshal execution payload: %v", err)
+	}
+
+	if createPayload.Metadata.Source == "" {
+		createPayload.Metadata.Source = "http://taskcluster-cli/task/create"
+	}
+
+	tags := make(map[string]string)
+	tags["misc_info"] = "created by taskcluster-cli"
```
djmitche: Probably better in the description :)

cmds/task/create.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	q := queue.New(creds)
+	resp, err := q.CreateTask(taskID, createPayload)
+	if err != nil {
+		return fmt.Errorf("could not create task: %v", err)
+	}
+
+	// If we got no error, that means the task was successfully created
+	fmt.Fprintf(cmd.OutOrStdout(), "Task %s created\n", resp.Status.TaskID)
```
djmitche: It would be sweet if this went on to poll the task until it becomes running, and then tails its output.

cmds/task/create.go
t0xicCode: Will remove.

cmds/task/create.go
t0xicCode: SchedulerId is the ID of the scheduler that defined the task, correct? I considered that as set because these tasks will always be defined by `taskcluster-cli`.

I'll see what I can do for the deadline and expires parameters

cmds/task/create.go
t0xicCode: I was thinking of adding that as a separate command. Or maybe adding an `--await` flag to do it, but making the "waiting for task completion" aspect optional.

cmds/task/create.go
djmitche: Yes, you're right about schedulerId -- I should have updated my review comment once I figured out this wasn't a generic "create task" function!

yannlandry:
@lteigrob help/flag and standard output (i.e. to confirm some operation was successful) messages are generally formatted with capitalization and punctuation, while error messages (passed through `error` objects) are lowercase and not punctuated, because they are formatted appropriately by cobra before being displayed. Are there messages that do not respect this format?

lteigrob:
@yannlandry Ah, okay. I'm good then :)

t0xicCode:
@yannlandry approval?

yannlandry:
Approved, but of course please get the checks to pass :smile:

PR #119, opened 2017-03-13
Add config command to the root command

djmitche:
The types were OK for this, but the command was being added to itself so it panicked when run.  `taskcluster version` has the same problem but I'll address it in #107.  Are there others?

### DIFF
--- a/cmds/config/get.go
+++ b/cmds/config/get.go
@@ -5,6 +5,7 @@ import (
 	"os"
 
 	"github.com/spf13/cobra"
+	"github.com/taskcluster/taskcluster-cli/cmds/root"
 )
 
 func init() {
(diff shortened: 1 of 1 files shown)


yannlandry:
Actually, that's not quite right. Your change adds the `get` command to the root command. However, the `get` command is a subcommand of `taskcluster config`, and that's how it was being registered first. The `Command` (without the `root` prefix) in package `configCmd` is the command `taskcluster config`.

What kind of panic did you get when trying to run the command?

djmitche:
Hm, you're right.  `taskcluster version` was definitely failing ("can't be a subcommand of itself" or something to that effect), but `taskcluster config get` works fine.  Sorry!

PR #121, opened 2017-03-15
Don't fetch subpackages in the first `go get`

djmitche:
@petemoore suggested this as a way to avoid unnecessary go-getting, since anything matched by the /... will be vendored, anyway.

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -21,7 +21,7 @@ tasks:
         - /bin/bash
         - '-c'
         - >-
-          go get -t github.com/taskcluster/taskcluster-cli/... 
+          go get -t github.com/taskcluster/taskcluster-cli
           && cd  /go/src/github.com/taskcluster/taskcluster-cli 
           && git fetch {{ event.head.repo.url }} {{ event.head.ref }} 
           && git checkout {{ event.head.sha }} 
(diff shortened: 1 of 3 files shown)


yannlandry:
Asked for review from @t0xicCode because he did some work on vendoring. I assume he'd have a say.

djmitche:
Indeed, that is why it failed!  I'll add testify.

lteigrob:
Will there be other dependencies we'll need to add as we keep going? 

I'm not sure if this would help, but this seems like a way to store the dependencies we need:
https://github.com/tools/godep
Then we might be able to avoid explicitly adding new ones to taskcluster.yml. There's use for vendoring as well, so we can also get around the duplicates problem with this. We'd have everything we need in one place. 

Perhaps for now just adding-in is fine. This might be a more permanent solution for a later time. ....assuming this is a solution at all

djmitche:
It looks like I had missed github.com/stretchr/testify/suite.  I only just learned that vendoring is per-package, not per repo.  The more you learn!

I think it makes sense to vendor things as we use them.  Leaving the `...` out, as the first commit does, means that we'll catch anything we forget to vendor, since it will cause the tests to fail instead of quietly being fetched into the gopath by the first `go get`.

I'm obviously not an expert, but I *think* godep is an alternative to govendor.  So I think continuing to use govendor here is the right approach.

djmitche:
Green tests!  @lteigrob @t0xicCode either of you want to give me a green checkmark? :)

lteigrob:
Okay, cool. Thanks for the explanation :)

PR #122, opened 2017-03-15
[NOT WORKING] upload releases

djmitche:
I'm trying to write a separate go binary that will be able to upload release files to S3 using `auth.awsS3Credentials`.

When I run this, I get
```
$ go run upload_release/upload.go                                                                                                                                                                                                    
# command-line-arguments
upload_release/upload.go:18: cannot use &creds (type *"github.com/taskcluster/taskcluster-cli/vendor/github.com/taskcluster/taskcluster-client-go".Credentials) as type *"github.com/taskcluster/taskcluster-client-go".Credentials in argument to auth.New
```
it's like half the time it finds the vendored client, and half the time it finds the client in my GOPATH.  I don't really care which one it uses, just be consistent..

### DIFF
--- a/upload_release/upload.go
+++ b/upload_release/upload.go
@@ -1,0 +1,64 @@
+package main
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/s3"
+	tcclient "github.com/taskcluster/taskcluster-client-go"
+	"github.com/taskcluster/taskcluster-client-go/auth"
+)
+
+func getSTSCredentials() (id, secret, token string, err error) {
+	// TODO: make sure this works with proxy
+	creds := tcclient.Credentials{}
+	client := auth.New(&creds)
+	resp, err := client.AwsS3Credentials("read-write", "downloads-taskcluster-net", "taskcluster-cli", "")
+	if err != nil {
+		return "", "", "", err
+	}
+	return resp.Credentials.AccessKeyID,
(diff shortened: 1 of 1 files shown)


petemoore:
> it's like half the time it finds the vendored client, and half the time it finds the client in my GOPATH. I don't really care which one it uses, just be consistent..

@djmitche That's because `github.com/taskcluster/taskcluster-client-go` is vendored, but `github.com/taskcluster/taskcluster-client-go/auth` [isn't](https://github.com/djmitche/taskcluster-cli/blob/6e46939a5fb5a177adcc9def383d7953facf7294/vendor/vendor.json#L101-L112). :smile:

You need to vendor the auth package with:

```
govendor add github.com/taskcluster/taskcluster-client-go/auth
```

petemoore:
Note, if you don't want the `upload_release` dependencies to be pulled down for taskcluster-cli developers not using that package, you could create a separate `upload_release/vendor` folder for just the `upload_release` dependencies that are not already in the main vendor folder. I don't know if `govendor` supports this design, but the go programming language certainly does.

See [spec](https://docs.google.com/document/d/1Bz5-UB7g2uPBdOx-rw5t9MxJwkfpx90cqG9AFL0JAYo/edit):

> If there is a source directory d/vendor, then, when compiling a source file within the subtree rooted at d, import "p" is interpreted as import "d/vendor/p" if that path names a directory containing at least one file with a name ending in “.go”.
>
> When there are multiple possible resolutions, the most specific (longest) path wins.
>
> The short form must always be used: no import path can contain “/vendor/” explicitly.
>
> Import comments are ignored in vendored packages.

As you see, since the _longest_ path wins, you can put `upload_release` specific packages in the  `upload_release/vendor` package, but fall back on packages in `/vendor`.

djmitche:
Thanks! I'll use this to finish up #49.

Issue #117, opened 2017-03-10
Error in `task log` when the public logs don't exist


t0xicCode:
When I try to watch or fetch the logs for a task that was just created (pending, unscheduled, or otherwise), I'm getting an ugly error message:

```
{
  "code": "ResourceNotFound",
  "message": "Artifact not found\n----\nmethod:     getLatestArtifact\nerrorCode:  ResourceNotFound\nstatusCode: 404\ntime:       2017-03-10T15:46:56.672Z",
  "requestInfo": {
    "method": "getLatestArtifact",
    "params": {
      "0": "public/logs/live.log",
      "taskId": "8aATqJZQSfihCp0cI0axdA",
      "name": "public/logs/live.log"
    },
    "payload": {},
    "time": "2017-03-10T15:46:56.672Z"
  }
}
Error: Received unexpected response code 404
```

The task log should catch that error (logs don't exists), and I think that it should first check the status of the task, so that it can display better error messages if the task doesn't exist or the logs haven't been created yet.

gregarndt:
I definitely like the idea of the error messages being clearer and checking task state.  If the task is currently defined or pending, then the run is not in a state that could have ever produced the log and an error could explain the task has not been run yet. nice suggestion!

PR #123, opened 2017-03-15
Add error handling for invalid task id's and tasks where the logs hav…

AyubMohamed:
…en't been created yet

For issue https://github.com/taskcluster/taskcluster-cli/issues/117


t0xicCode:
@AyubMohamed you should investigate the cause of the segfault that all tests experienced before actually merging this.

AyubMohamed:
@t0xicCode Should be fixed now, the segfault was due to not passing in credentials during the test so taskcluster-client complained. This is why we have tests right!

t0xicCode:
:+1: :shipit:

PR #129, opened 2017-03-24
Status command

yasch007:
This is the `status` command I have been working on. I hope it's good enough for a v.1 release. 

Tested on Linux, Mac and Windows.

Since client auto-uptdates are not implemented yet, the `status` command takes care of refreshing the list of services from `http://references.taskcluster.net/manifest.json` at most once per 24 hours (and caches the required metadata).

Colourised output included.


djmitche:
Can you update to resolve the conflicts?

I'd like to get review from one of the other people working on this project - @AyubMohamed, do you have a bit of time perhaps?

yasch007:
@djmitche Many thanks! Merge conflict resolved.

cmds/root/command.go
```diff
@@ -5,7 +5,7 @@ import "github.com/spf13/cobra"
 var (
 	// Command is the root of the command tree.
 	Command = &cobra.Command{
-		Use:   "taskcluster",
+		Use:   "taskcluster-cli",
```
djmitche: The binary is named `taskcluster`, not `taskcluster-cli`

cmds/status/status.go
```diff
@@ -0,0 +1,267 @@
...
+func CacheFilePath() string {
+	// 1. find out where home directory is (panic in case of error)
+	usr, err := user.Current()
+	if err != nil {
+		panic(err)
+	}
+	home := usr.HomeDir
+
+	// 2. return file path <home directory>/ .taskcluster-cli/ cmds/ status / cache.json
+	return filepath.Join(home, ".taskcluster-cli", "cmds", "status", "cache.json")
```
djmitche: This should use the `XDG_*` env vars just like the config package does.

petemoore:
> Downloading the list of endpoints from references is an interesting idea, but it is contrary to the existing practice of generating code based on that reference, checking in the code, and making a new release of the tool. I don't think we should solve the same problem twice in different ways -- rather, I think that the automatic code generation process should also provide enough information for this command to operate.

@djmitche I suggested @yasch007 could implement this way, since the binary is not yet self-updating. Thus, currently if the endpoints change, the clients won't pick this up (and since Yannick leaves in a week I thought best not to depend on a feature that isn't implemented yet).

> I'm not sure I see the benefit of caching this information, either: if we have the code to fetch it, why not just fetch it on every run?

In order for the `taskcluster --help` command to be able to list the possible services that can be queried, this information needs to be available on package initialisation. Thus we decided it would be a bad idea to run the same network calls every time any command was run. Caching overcomes this issue, and if a user requires an updated service list immediately (and doesn't want to wait for cache to expire) they can of course manually delete the cache.

cmds/status/status.go
petemoore: @djmitche Technically, this is not config (it should not be editable by a human). It is variable data, so doesn't belong in a config directory - but possibly could go in e.g. `/var`.

djmitche:
Yes, but XDG defines dirs for caches, too.

So we already use code generation (apis) and external code generation (tc-client-go).  I really don't want to add a third method of getting information about the APIs.  So we could potentially have the same tool using three different representations of the services (ping for service foo appears in `taskcluster status`, and a new version of its API is used in the implementation of the `taskcluster foo` subcommand, but `taskcluster api foo` has an older version of the API that lacks some fancy new parameter. 

Perhaps we can think of the code-generation approach as a form of caching?

djmitche:
Pete and I talked a bit more about the design here.  What we came around to is, it's probably sensible for everything (the `taskcluster api ..` commands, too) to be downloading schema information "live" like this.  But that would be a whole bunch of additional work that is way out of scope for this PR.

So I think we can merge this PR, with the minor fixes above (using the same technique to find cache directories as we do to find config directories, and not renaming the command from `taskcluster` to `taskcluster-cli`).

So let's get this fixed up, and if there's time a next step might be to think about making a module to handle finding config and cache dirs, so we have it in one place (I'm sure it'll be different on Windows, after all!).  Then if another subcommand wants to do some caching, it can re-use that directory.

yasch007:
@djmitche  I have addressed your comments, it should be ready for a second review now.  Many thanks!

djmitche:
@yasch007 by the way, taskcluster-bot wouldn't complain so much if you'd accept the invitation to the repository :)

vendor/vendor.json
```diff
@@ -71,6 +71,12 @@
 			"revisionTime": "2016-01-10T10:55:54Z"
 		},
 		{
+			"checksumSHA1": "uPCqYCT45y5/tdWgMJgShUTJ+y0=",
+			"path": "github.com/shibukawa/configdir",
```
djmitche: neat!

If you can find time this week, do you thin you could update `config/serialization.go` to use github.com/shibukawa/configdir, too (in a separate PR)?  If not, please do open a Github issue on the topic so we don't forget.

cmds/status/status.go
```diff
@@ -53,16 +53,10 @@ type (
...
-		panic(err)
-	}
-	home := usr.HomeDir
-
-	// 2. return file path <home directory>/ .taskcluster-cli/ cmds/ status / cache.json
-	return filepath.Join(home, ".taskcluster-cli", "cmds", "status", "cache.json")
+func Cache() (cache *configdir.Config) {
+	configDirs := configdir.New("taskcluster", "taskcluster-cli")
+	cache = configDirs.QueryCacheFolder()
+	return
```
djmitche: I think this can be omitted, since it is the last line in the function

cmds/status/status.go
```diff
@@ -144,13 +138,18 @@ func ReadCachedURLsFile(path string) (cachedURLs *CachedURLs, err error) {
...
 // already, and creating parent folders, if required), using the current time
 // for the retrieval timestamp.
-func (p PingURLs) Cache(path string) (cachedURLs *CachedURLs, err error) {
-	color.Magenta("Writing cache file %v", path)
-	parentDir := filepath.Dir(path)
+func (p PingURLs) Cache(cache *configdir.Config, cachePath string) (cachedURLs *CachedURLs, err error) {
+	color.Magenta("Writing cache file %v", filepath.Join(cache.Path, cachePath))
+
+	///////////////////////////////////////////////////////////////////////
+	// workaround until https://github.com/shibukawa/configdir/pull/2 lands
```
djmitche: Please add a github issue in this repo to remove this workaround when the time is right.

vendor/vendor.json
yasch007: I have created #135. If I get the chance to add it this week I will.

cmds/status/status.go
yasch007: I have created #137.

petemoore: Fixed in #142

Issue #126, opened 2017-03-20
New feature: Cancel a subset of a group


JohanLorenzo:
This need came out last Friday, when releng had several releases in flight and all of them were claiming the same workers. One way to solve this issue is to [handle priorities in a different manner](https://bugzilla.mozilla.org/show_bug.cgi?id=1348773). Nonetheless, having a manual way to cancel/rerun a certain type of tasks, would be great to manually help a graph to claim workers.

For example:
```sh
taskcluster group cancel --worker-type 'signing-worker-v1' $TASK_GROUP_ID
```
will:
1. look for unscheduled/pending/running tasks that match this worker type
1. display the name of the tasks found
1. ask for a confirmation
1. if confirmation given, cancel them

`rerun` may also follow the same workflow.

What do you guys think?

djmitche:
I like this!  @yannlandry do you want to jump on this one?

yannlandry:
Yep, time to jump on task management!

escapewindow:
In this workflow, we want `rerun` to only rerun the previously cancelled jobs.

e.g.
- we have task1 and task2, both blocking downstream tasks
- task1 is completed, and downstream tasks have also completed
- task2 is cancelled, and downstream tasks are pending

If we rerun task2, all is as expected: task2 completes for the first time, and downstream tasks are unblocked.

If we rerun task1, completed downstream tasks don't necessarily retrigger, and we may end up with indexes and routes with differing artifacts+hashes than expected.

PR #131, opened 2017-03-25
Add cancellation by worker-type, and cancellation confirmation

yannlandry:
This is to address issue #126. We might not want to merge it right now.

It builds, but I don't really know anything about TaskCluster itself, and I'm not sure how to go about testing this (even manually).

Otherwise, I've added code that enables to filter tasks to cancel with extra criteria (for now, only `worker-type` has been implemented) and code that asks for confirmation before cancelling tasks.

Let me know what you think.

### DIFF
--- a/cmds/group/actors.go
+++ b/cmds/group/actors.go
@@ -52,7 +52,7 @@ func runCancel(credentials *tcclient.Credentials, args []string, out io.Writer, 
 	}
 
 	// ask for confirmation before cancellation
-	if confirmCancellation(tasks, out) == false {
+	if !confirmCancellation(tasks, out) {
 		fmt.Fprintln(out, "Cancellation of tasks aborted.")
 		return nil
 	}
--- a/cmds/group/actors.go
+++ b/cmds/group/actors.go
@@ -11,6 +11,17 @@ import (
 	"golang.org/x/net/context"
 )
 
+func init() {
+	cancelCmd := &cobra.Command{
+		Use:   "cancel <taskGroupId>",
+		Short: "Cancel a whole group by taskGroupId.",
+		RunE:  executeHelperE(runCancel),
+	}
+	cancelCmd.Flags().StringP("worker-type", "w", "", "Only cancel tasks with a certain worker type.")
+
+	Command.AddCommand(cancelCmd)
+}
+
 // runCancel cancels all tasks of a group.
 //
 // It first fetches the list of all tasks associated with the given group,
--- a/cmds/group/actors.go
+++ b/cmds/group/actors.go
@@ -17,30 +17,44 @@ import (
 // then filters for only cancellable tasks (unscheduled, pending, running),
 // and finally runs all cancellations concurrently, because they are
 // independent of each other.
-func runCancel(credentials *tcclient.Credentials, args []string, out io.Writer, _ *pflag.FlagSet) error {
+func runCancel(credentials *tcclient.Credentials, args []string, out io.Writer, flags *pflag.FlagSet) error {
 	q := queue.New(credentials)
 	groupID := args[0]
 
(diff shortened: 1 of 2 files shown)


yannlandry:
Will merge once I get confirmation from @johanlorenzo.

cmds/group/group.go
```diff
@@ -19,6 +19,7 @@ func init() {
 		Short: "Cancel a whole group by taskGroupId.",
 		RunE:  executeHelperE(runCancel),
 	}
+	cancelCmd.Flags().StringP("worker-type", "w", "", "Only delete tasks with a certain worker type.")
```
JohanLorenzo: Nit: I don't think we can delete task. "only cancel" sounds more appropriate.

cmds/group/actors.go
```diff
@@ -103,3 +117,48 @@ func runCancel(credentials *tcclient.Credentials, args []string, out io.Writer,
...
+	return true
+}
+
+// confirmCancellation lists the tasks to be cancelled and prompts to confirm cancellation
+func confirmCancellation(tasks []string, out io.Writer) bool {
+	// list tasks
+	fmt.Fprintf(out, "The following %d tasks will be cancelled:\n", len(tasks))
+
+	for _, task := range tasks {
+		fmt.Fprintf(out, "\tTask %s\n", task)
```
JohanLorenzo: Feature request: Would it be possible to print out the name as well? That'll help in diagnosing whether the worker-type is correct.

cmds/group/actors.go
```diff
@@ -103,3 +117,48 @@ func runCancel(credentials *tcclient.Credentials, args []string, out io.Writer,
...
+	}
+
+	for {
+		fmt.Fprint(out, "Are you sure you want to cancel these tasks? [Y/n] ")
```
JohanLorenzo: I like the fact "yes" is not the default choice 😃 

Nit: I think the choice in upper case is generally the default choice. I'd put "Y" in lower case, then

cmds/group/actors.go
djmitche: (that's `Task.Metadata.Name`)

cmds/group/actors.go
yannlandry: There is in fact no default choice :wink: 

The command reloops until you enter **Y** or **N**... so for now I'll put `[y/n]`.

cmds/group/group.go
yannlandry: Good point.

yannlandry:
Fixed all the nits. @JohanLorenzo please confirm that everything is still working "in real life", and we can merge then.

JohanLorenzo:
I'm sorry. Due to conflicts in `cmds/group/actors.go, I couldn't test out the latest changes.

yannlandry:
Yes, there are conflicts between this branch and the `master` branch. Can you simply pull `cancel-group-subset` and test that branch?

yannlandry:
As per the code, it also catches running tasks, so maybe it does what you already want.

Anyhow, I'll resolve the conflicts and merge.

djmitche:
I'm sorry, I said the wrong thing -- we might want it to cancel pending jobs as well.  But I was right about changing that in a followup :)

djmitche:
Interesting, this ran out of memory in Travis on master.

yannlandry:
Yes, I broke the test and somehow it didn't come up until I merged into master. Fix is in PR #149.

t0xicCode:
That's why my computer crashed. Don't run tests right before you're due to
present!

JohanLorenzo:
On the bright side, I retested the patch IRL, and it worked as expected 😃 . Tested scenarios:

* Cancel a graph entirely with pending, running and unscheduled
* Cancel a given worker type which were unscheduled. 

Thanks again @yannlandry

PR #133, opened 2017-03-27
[NOT TO MERGE YET] Serialization.go testing

lteigrob:
I've just got a basic test to get things going, and have ideas from Yann on how to work on that.

I made configFile() return both a string and io.Reader since I need the io.Reader as an input for Load but elsewhere a string return is needed. Any ideas for a less clunky way to deal with this change would be welcome... readers are still weird for me ( @walac I know you had a way to do this, but I didn't see all the specifics )

### DIFF
--- a/config/serialization.go
+++ b/config/serialization.go
@@ -16,7 +16,7 @@ import (
 )
 
 // configFile is the location of the configuration file
-func configFile() io.Reader {
+func configFile() (io.Reader, string) {
 	configFolder := os.Getenv("XDG_CONFIG_HOME")
 	if configFolder == "" {
 		homeFolder := os.Getenv("HOME")
@@ -27,8 +27,8 @@ func configFile() io.Reader {
 			configFolder = filepath.Join(homeFolder, ".config")
 		}
 	}
-	returnConfigLoc := filepath.Join(configFolder, "taskcluster.yml")
-	return strings.NewReader(returnConfigLoc)
+	configFileStringString := filepath.Join(configFolder, "taskcluster.yml")
+	return strings.NewReader(configFileStringString), configFileStringString
 }
 
 // Load will load confiration file, and initialize a default configuration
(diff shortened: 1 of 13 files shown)


config/serialization.go
```diff
@@ -24,26 +27,30 @@ func configFile() string {
...
 	config := make(map[string]map[string]interface{})
 
 	// Read config file and unmarshal into config overwriting default values
 	// if ioutil.ReadFile returns an error, it means the config file couldn't
 	// be found and we just skip
-	configFile := configFile()
-	if data, err := ioutil.ReadFile(configFile); err == nil {
+	buf := new(bytes.Buffer)
+	buf.ReadFrom(configFileReader)
+	configFileString := buf.String()
```
walac: I don't get the role of `buf`

config/serialization.go
```diff
@@ -169,7 +176,7 @@ func Save(config map[string]map[string]interface{}) error {
 	}
 
 	// Write config file
-	configFile := configFile()
+	_ ,configFile := configFile()
```
walac: If this is the spot you were talking about requiring filename, you can make the config write function receive an `io.Writer` and make `configFile` return `io.ReadWriterCloser`

config/serialization_test.go
```diff
@@ -0,0 +1,32 @@
...
+    }
+    RegisterOptions("cmd", options)
+    r := strings.NewReader(configSample)
+    m, err := Load(r)
+    assert.NoError(t, err)
+    _, ok := m["cmd"]
+    assert.True(t, ok)
+    _, ok = m["cmd"]["foo"]
+    assert.True(t, ok)
+
```
yannlandry: Let's remove this empty line and instead make some space above. Maybe a couple really short comments so we can follow flow easily, especially useful for when code gets complex.

config/serialization.go
lteigrob: Ah okay. Then the places where I was using the string return would use io.writer instead... So I could pare down my configFile return to only io.ReadWriterCloser instead of both io.Reader and string, is that right?

config/serialization_test.go
lteigrob: Yes, this part will definitely be getting a lot of work still... I'll do those suggestions

config/serialization.go
```diff
@@ -24,26 +27,30 @@ func configFile() string {
...
+	return strings.NewReader(configFileString), configFileString
 }
 
 // Load will load confiration file, and initialize a default configuration
 // if no configuration is present. This only returns an error if a configuration
 // file is present, but we are unable to parse it.
 // TODO	we could simplify this function and only go through the definitions once
 //		and what happens if a value is tied to an env var AND to the config file?
-func Load() (map[string]map[string]interface{}, error) {
+func Load(configFileReader io.Reader) (map[string]map[string]interface{}, error) {
```
lteigrob: So I would separate out configFile() into two functions, one that returns the string location and one that returns the io.Reader. Then in init.go and other calls for Load I can lose that extra step of extracting the io.Reader that I need. That's awesome. 

But then if I revert to Load without parameters, what would my testing look like, without being able to call from my test file and pass in the fake config data? I'd originally tried something like this using an interface, but discussing with @walac, this seemed the way to go instead.

lteigrob:
@yannlandry, I believe my last commit follows your suggestions, and does seem to simplify things. Should I really get rid of the map parameter for Save as well? It wasn't one I had added but wouldn't be difficult to lose. 

I'm also wanting to fill out my ToDo list in the test file, then I think I'll be done<ish>. I've done some searches on what I want to do, but haven't quite found what I want. Can you (or anyone) point me in the right direction for this?

config/serialization_test.go
```diff
@@ -0,0 +1,46 @@
...
+                         `
+    options := map[string]OptionDefinition{
+       "foo": OptionDefinition{},
+    }
+    
+    // Register the fake config options
+    RegisterOptions("cmd", options)
+    
+    // Call Load() with the test data; reassign configReader
+    configReader,_ = os.Create(configSample)
```
walac: `os.Create` receives the file name, not the file content.

lteigrob: Oh right, okay thanks!

djmitche:
I think this has probably bitrotted too much to be revived :(

PR #134, opened 2017-03-28
Avoid `go get` in travis build

petemoore:
Running `go get` will pull in package dependencies that are not vendored, so if we avoid using `go get` in our travis builds, we'll be sure we've vendored all our build and test dependencies, since there is no chance for libraries to make it into the go path via another means.


petemoore:
There you go. :-)

djmitche:
Thanks!

Issue #140, opened 2017-03-29
Golinter warnings on status.go


AyubMohamed:
I believe @yannlandry worked on status.go, there seems to be some warnings with golinter on `status.go`. 

````
cmds/status/status.go:221:2:warning: 'if err != nil { return err }; return nil' can be simplified to 'return err' (S1013) (gosimple)
cmds/status/status.go:246:5:warning: should omit comparison to bool constant, can be simplified to servstat.Alive (S1002) (gosimple)
cmds/status/status.go:32:2:warning: exported type PingURLs should have comment or be unexported (golint)
cmds/status/status.go:34:2:warning: exported type CachedURLs should have comment or be unexported (golint)
cmds/status/status.go:39:2:warning: exported type PingResponse should have comment or be unexported (golint)
cmds/status/status.go:44:2:warning: exported type API should have comment or be unexported (golint)
cmds/status/status.go:49:2:warning: exported type APIEntry should have comment or be unexported (golint)
cmds/status/status.go:55:1:warning: comment on exported function Cache should be of the form "Cache ..." (golint)
cmds/status/status.go:166:1:warning: exported method CachedURLs.Expired should have comment or be unexported (golint)
cmds/status/status.go:174:1:warning: comment on exported function ScrapePingURLs should be of the form "ScrapePingURLs ..." (golint)
cmds/status/status.go:210:6:warning: func objectFromJsonURL should be objectFromJSONURL (golint)
```

djmitche:
It was @yasch007 not @yannlandry 

@yasch007 this didn't run on your PR because you're not a member of the repo.  I think we can change that to allow taskcluster jobs to run on pull requests from anyone.  Anyway, if you can fix up these lint issues, I'd appreciate it.

AyubMohamed:
Woops, sorry about that.

PR #141, opened 2017-03-29
allow PRs from public

djmitche:
This should allow people who are not repository collaborators to run tasks.  We saw this in #126 where @yasch007's PR wasn't tested in taskcluster.

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -1,4 +1,5 @@
 version: 0
+allowPullRequests: public
 metadata:
   name: "TaskCluster .taskcluster.yml file"
   description: "Tasks for taskcluster-cli"


lteigrob:
In the taskcluster.yml task log I'm only seeing warnings to do with gosimple and golint from status.go that are causing this to fail. I see issue #140 should fix this. 

But travis.yml is failing for fetchers_test.go's TestLogCommand instead.  It's just a test of the function, with fake data passed in, so should be pretty isolated... @AyubMohamed, do you know why that would pop up now?

djmitche:
Sure, no problem.  The "allow edits from maintainers" button is checked, so you should be able to modify it.

djmitche:
rebased!

djmitche:
*shakes fist at golint*

I feel like golint randomly finds new problems every few days, so our PRs are always failing..

lteigrob:
Haha, yeah. My local 'go test' will be fine and then I push and everything is terrible. Glad it's not just me.

djmitche:
I feel like we're spending way too much time sweating lint errors instead of doing production work :(

@lteigrob probably one thing that would help is running taskcluster jobs on pushes to master, too -- currently they only run on pull requests.  The last pull reqest to be merged also failed (but, with different lint errors.. https://tools.taskcluster.net/task-group-inspector/#/e4WPJRJeSdaSdKxeWzDlNQ/e4WPJRJeSdaSdKxeWzDlNQ?_k=1k0vn1).

That should be a matter of adding 'push' to task.extra.github.events.

Once that's in place, let's get a pull request that addresses all of the gometalinter errors and is green, then merge it.  After that, if we see more lint appear out of nowhere, we can decide to just disable the linter or figure out why it's being capricious and fix it.

djmitche:
Oh, and while we're waiting, maybe we can land this patch?

djmitche:
Thanks :)

PR #147, opened 2017-04-07
Make lint checks actually pass, and easier to run

djmitche:
Just about every pull request since we have started running golint has has a big red X because golint found some lint problem in code the PR hadn't changed, but golint had not found that error in the last commit on master.  That means we're digging through logs to find out whether the red X is "real" or "fake", and we're tempted to ignore it.

I don't know why golint is doing this -- it's possible it's just under active development, and gometalinter is pulling in the latest, and that's adding new errors every few days.  At any rate, I think it's a net negative on development performance on this project.

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -14,6 +14,42 @@ tasks:
           - pull_request.opened
           - pull_request.synchronize
           - pull_request.reopened
+          - push
+    payload:
+      maxRunTime: 3600
+      image: 'golang:1.8'
+      command:
+        - /bin/bash
+        - '-c'
+        - >-
+          mkdir -p  /go/src/github.com/taskcluster/taskcluster-cli
+          && cd  /go/src/github.com/taskcluster/taskcluster-cli
+          && git init
+          && git fetch {{ event.head.repo.url }} {{ event.head.ref }}
+          && git checkout {{ event.head.sha }}
+          && go get -u github.com/alecthomas/gometalinter
+          && gometalinter --install --force
+          && go install ./...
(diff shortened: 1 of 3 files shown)


t0xicCode:
We should still recommend that contributors run it before pushing though.

djmitche:
I'm not sure how we could fairly do that without running into the same issues, though.

djmitche:
Hm, actually we (still?) don't have lint running on pushes to the master branch, so it's possible we just keep landing worse lint.

djmitche:
..but when I run the gometalinter command from `.taskcluster.yml`, I get 715 errors, which is a LOT more than we've been seeing in pushes.

t0xicCode:
Seems like all of the errors are due to `gotype`. Disabling that one seem to fix it running locally.

djmitche:
Indeed!  I did some other refactoring, too.. update incoming..

djmitche:
This is incredibly frustrating.  I ran gometalinter before that push, and it ran without errors.  Pushed, updated the PR, and ran again just for fun.  And
```
cmds/shell/cmd.go:85:21:warning: error strings should not be capitalized or end with punctuation or a newline (golint)
cmds/shell/cmd.go:93:21:warning: error strings should not be capitalized or end with punctuation or a newline (golint)
```
It's correct about that issue, but if it's not going to detect the issue 100% of the time, it's no good to us.

djmitche:
Oh, it didn't find it earlier because I'd removed `--enable=gometalinter` in that command line.  I shouldn't jump to conclusions :)

PR #150, opened 2017-04-10
Move config get back in config

t0xicCode:
Seems like `config get` got put into the root instead of `config`. This fixes that :)


djmitche:
(I rebased)

t0xicCode:
thanks!

PR #151, opened 2017-04-10
Run tests and lint on push, but only to master

djmitche:


djmitche:
OK, now we can make a release :)

t0xicCode:
I reverted my tags

t0xicCode:
(I jumped the gun slightly). All fixed now :)

PR #153, opened 2017-04-10
link to the latest releases

djmitche:
Looking in the S3 bucket, these are all named `taskcluster` and not e.g. `taskcluster-windows-amd64.exe`.  I kinda like the lack of platform in the name, but the missing `.exe` isn't so good.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,8 +21,16 @@ or `--help` parameter to get a command's help information.
 To install, download the `taskcluster` binary for the latest release your
 platform, and run it!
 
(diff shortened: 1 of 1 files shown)


djmitche:
I can take a look at the upload if that wasn't what you intended -- let me know.

djmitche:
Hm, they are also all 0-length files.  That seems wrong :(

djmitche:
And let me be honest, uploads.taskcluster.net was not a good idea.  We should be using the index.

PR #157, opened 2017-04-14
make releases via the index

djmitche:
This only uses `latest` for the moment, because from what I can tell tc-github doesn't actually provide the release name as a useful value.  I'll make a followup, possibly after fixing that, once I see this much work.

@t0xicCode fyi


.taskcluster.yml
```diff
@@ -95,7 +96,11 @@ tasks:
           && cd  /go/src/github.com/taskcluster/taskcluster-cli
           && git fetch {{ event.head.repo.url }} {{ event.version }}
           && git checkout FETCH_HEAD
-          && make prep release upload
+          && make prep release
+      artifacts:
+        public:
+          path: /go/src/github.com/taskcluster/taskcluster-cli/taskcluster/build
+          type: directory
```
jonasfj: directory is not a good idea... Use file artifacts (in some soon to be future, we'll have workers that fails tasks if artifacts are missing, if we didn't already fix that)...

Also it force a consistent naming...

djmitche: .. but it's redundant, and likely to be forgotten when adding a new platform, in which case we'd have missing artifacts anyway (unless the worker is using AI to say "I bet you meant this file too")

djmitche:
@t0xicCode it does not.  We could add a route for that, I think -- that's what I mentioned in the original comment.

PR #163, opened 2017-08-30
Add a task definition fetching command

imbstack:
Forgive me for whatever golang sins I've committed. I want to have this for doing things like fetching the tasks from [this list](https://pastebin.mozilla.org/9031069) and passing the results to `jq` and friends.

### DIFF
--- a/cmds/task/fetchers.go
+++ b/cmds/task/fetchers.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"encoding/json"
 
 	"github.com/spf13/pflag"
 	tcclient "github.com/taskcluster/taskcluster-client-go"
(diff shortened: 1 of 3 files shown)


vendor/vendor.json
```diff
@@ -383,7 +383,7 @@
 			"revisionTime": "2016-12-29T22:47:41Z"
 		},
 		{
-			"checksumSHA1": "oBjdFvDUZpioGsB6URqe16hDQUg=",
```
djmitche: any idea why these checksums changed?

imbstack: Not exactly sure. I'll remove it from the patch.

imbstack:
Oh, I didn't figure that out about the `api`  subcommand 😭

PR #165, opened 2017-10-02
Add support for bash completion file production

srfraser:
Use cobra's own tools for generating the shell script.

    ./taskcluster completions
    source bash_completion.sh

Or:

    ./taskcluster completions mycompletion.sh
    source mycompletion.sh

or moving the file into `/etc/bash_completion.d` on Linux and resetting the shell.

### DIFF
--- a/cmds/completions/completions.go
+++ b/cmds/completions/completions.go
@@ -1,0 +1,31 @@
+package completions
+
+import (
+	"github.com/spf13/cobra"
+	"github.com/taskcluster/taskcluster-cli/cmds/root"
+)
+
+var (
+	defaultFilename = "bash_completion.sh"
+)
+
+func init() {
+	// Add the task subtree to the root.
+	use := "completions <filename (default:" + defaultFilename + ")>"
+	completionsCommand := &cobra.Command{
+		Short: "Provides bash completion script.",
+		RunE:  genCompletion,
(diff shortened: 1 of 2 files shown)


srfraser:
Example script at https://gist.github.com/srfraser/fb05cf2041fe28f004cd997d50c5d470

cmds/completions/completions.go
```diff
@@ -0,0 +1,31 @@
...
+)
+
+func init() {
+	// Add the task subtree to the root.
+	use := "completions <filename (default:" + defaultFilename + ")>"
+	completionsCommand := &cobra.Command{
+		Short: "Provides bash completion script.",
+		RunE:  genCompletion,
+		Use:   use,
+	}
```
jonasfj: So could I do:
`$ sudo taskcluster completions /etc/bash_completion.d/taskcluster.sh`

to install these completions globally?

Maybe we could put that in the `--help` message, or even make the default filename on linux... :)

srfraser: I didn't necessarily want to make the default output file one you needed special rights to write to, but adding the instructions to the help sounds like a good idea. Macs are... special

jonasfj: oh, yeah, good point..

PR #166, opened 2017-10-05
Added task await command

t0xicCode:
Added an `task await` command that sleeps (60 seconds at a time by default) until the task completes.

### DIFF
--- a/cmds/task/fetchers.go
+++ b/cmds/task/fetchers.go
@@ -187,6 +187,11 @@ func runAwait(credentials *tcclient.Credentials, args []string, out io.Writer, f
 	q := makeQueue(credentials)
 	taskID := args[0]
 
+	delay, _ := flagSet.GetInt("sleep")
+
+	fmt.Fprintf(out, "Waiting for completion of task %s\n", taskID)
+	fmt.Fprintf(out, "Checking every %v seconds\n", delay)
+
 	for {
 		s, err := q.Status(taskID)
 		if err != nil {
@@ -195,9 +200,10 @@ func runAwait(credentials *tcclient.Credentials, args []string, out io.Writer, f
 
 		switch s.Status.State {
 		case "completed", "failed", "exception":
+			fmt.Fprintln(out, getRunStatusString(s.Status.Runs[len(s.Status.Runs)-1].State, s.Status.Runs[len(s.Status.Runs)-1].ReasonResolved))
 			return nil
 		default:
-			time.Sleep(60 * time.Second)
+			time.Sleep(time.Duration(delay) * time.Second)
 		}
 	}
 }
--- a/cmds/task/fetchers.go
+++ b/cmds/task/fetchers.go
@@ -3,10 +3,11 @@ package task
 import (
 	"bufio"
 	"bytes"
+	"encoding/json"
 	"fmt"
 	"io"
 	"net/http"
-	"encoding/json"
+	"time"
 
 	"github.com/spf13/pflag"
 	tcclient "github.com/taskcluster/taskcluster-client-go"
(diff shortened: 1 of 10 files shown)


cmds/task/fetchers.go
```diff
@@ -180,3 +181,29 @@ func runLog(credentials *tcclient.Credentials, args []string, out io.Writer, fla
 
 	return nil
 }
+
+// runAwait gets the status of run(s) of a given task.
+func runAwait(credentials *tcclient.Credentials, args []string, out io.Writer, flagSet *pflag.FlagSet) error {
```
jonasfj: So this keeps polling the queue until the task is resolved..
I can see how it's useful, for small scripts or hacks where you want to wait for some thing to happen...

However, we don't have to be polling we have two other options:
 * Acquire pulse credentials from https://pulseguardian.mozilla.org/ and listen for events...
 * Implement a golang client for taskcluster-events (see: https://github.com/taskcluster/taskcluster-events)

For a quick update on what pulse is, take a look at:
https://wiki.mozilla.org/Auto-tools/Projects/Pulse
In short:
 * RabbitMQ is a pub-sub system, where you can publish messages and subscribe to messages,
 * Pulse is a RabbitMQ deployment with special rules, like:
   * We use topic exchanges meaning each message has a routing key on the form: "<key>.<words>.<...>" (dot separated)
   * All messages are considered public
   * Event publishers can only write to their own exchanges (anyone can read messages)
   * Access requires credential but these are easy to obtain
 * taskcluster-events is a simple gateway that allows for listening for pulse messages over a websocket without pulse credentials (so as to lower the entry bar, and be usable on the web)

It's a simple protocol, the idea is as follow:
1. Connect to `wss://events.taskcluster.net/v1//listen/websocket` using a websocket
2. Wait for message `{method: 'ready'}` from websocket
3. Send message `{method: 'bind', id: '<slugid.Nice()>', options: {exchange: '...', routingKeyPattern: '...'}}`
4. Wait for message `{method: 'bound', id: '<id-as-was-sent>'}`
5. Reload state of task from queue
6. Wait for message to arrive from websocket

If disconnected start from 1 again, as soon as you see the state of the task change to what you want, you're done...
You'll probably have to bind to each of the exchanges for task-completed, task-failed, etc... Notice that the routingKeyPattern could be: `primary.<taskId>.#` and then you'll only get message for said taskId (this is nice).

Information on pulse: https://wiki.mozilla.org/Auto-tools/Projects/Pulse
See queue exchanges: https://docs.taskcluster.net/reference/platform/taskcluster-queue/references/events

Implementations of client side for `taskcluster-events`: (yes, we have two)
* https://github.com/taskcluster/taskcluster-client-web/blob/master/src/WebListener.js
* https://github.com/taskcluster/taskcluster-client/blob/master/lib/weblistener.js

Server side implementation of `taskcluster-events`:
https://github.com/taskcluster/taskcluster-events/blob/master/events/socket.js

Note, if you wanted to do this, it would probably be smart to start a golang package (or subpackage) that focuses exclusively on implementing the `WebListener` (notice that this doesn't require any credentials to use).

t0xicCode: Sounds good. I threw this fairly quickly just to get it out. I'll look at it in more details after I land shell v1 :)

djmitche:
I wouldn't mind putting this version out there to see how it's used.  There's the risk that it will find wide use in some silly way and pound the queue, but .. there's a risk someone does the same already.

jonasfj:
That's a fair point, we can always optimize the implementation later...

--
Regards Jonas Finnemann Jensen.

2017-10-09 8:39 GMT-07:00 Dustin J. Mitchell <notifications@github.com>:

djmitche:
@jonasfj should we rebase this and merge, or abandon?

jonasfj:
I don't know :)

If there is a use-case for it we can merge it, but polling at an interval isn't a good way to do this... It's slow and adds a lot of overhead to the queue..

djmitche:
OK, I'll count that as "abandon".. sorry @t0xicCode !

t0xicCode:
No worries. I'll mark #81 as completed and we can re-open/create a new issue if the use-case comes up from releng or whomever.

Issue #54, opened 2017-01-11
Connect user's shell to a one-click-loaner


djmitche:
We have a fancy one-click-loaner option that is accessible via https://tools.taskcluster.net, but the terminal emulation is TERRIBLE and you can't copy/paste or do any of the normal terminal things.

Much better would be to connect my existing terminal to a one-click loaner:

```
dustin@lamport ~ $ taskcluster terminal $TASKID
# <--- shell prompt from loaner
```

This should support both websocket protocols
* https://github.com/taskcluster/docker-exec-websocket-server
* https://github.com/taskcluster/taskcluster-worker/tree/master/plugins/interactive/shellclient

jonasfj:
taskcluster-worker already has a shell command implemented here:
https://github.com/taskcluster/taskcluster-worker/tree/master/commands/shell

That sets up the shell to work in raw mode, and tears it down when disconnecting.
Granted it might still have some edge cases around cleanup if disconnected incorrectly or so, I don't remember.

Obviously the shellclient can be reused from taskcluster-worker, for the docker-exec-websocket-server needs a client library reimplemented in golang.

djmitche:
We have at least one user interested in this!

https://bugzilla.mozilla.org/show_bug.cgi?id=1335874

jonasfj:
How interactive shell works right now:
 1. User creates a task with `task.payload.features.interactive = true`
 2. Worker starts task and listens for websocket connections to:
    `https://<hostname>:<port>/<secret>/shell`
 3. Worker publishes a redirect artifact called `private/docker/shell.html`
    which redirects to: 
    `https://tools.taskcluster.net/shell/?v=1&socketUrl=https://<hostname>:<port>/<secret>/shell`
 4. User clicks on the `private/docker/shell.html` artifact and is redirected to:
    `https://tools.taskcluster.net/shell/?v=1&socketUrl=https://<hostname>:<port>/<secret>/shell`
 5. Users browser reads `v` and `socketUrl=https://<hostname>:<port>/<secret>/shell`
    from querystring. If `v=1` the browser uses `docker-exec-websocket-server` to
    open a shell over a websocket. If `v=2` it uses `ws-shell` (to do the same thing).
 6. Users browser exposes the shell object to the user through a terminal emulator
    running in the browser.

How to make this work in taskcluster-cli:
 1. User creates a task with `task.payload.features.interactive = true`
 2. Worker starts task and listens for websocket connections to:
[...]

PR #167, opened 2017-10-05
[WIP] #54 Shell v1 client

t0xicCode:
This is a WIP of the v1 shell client and protocol in go.


external/shellclient/v1/messages.go
```diff
@@ -0,0 +1,27 @@
+package v1
+
+const (
```
jonasfj: Maybe put this into a subpackage called `pkg/docker_exec_ws/const` since, there is no reason they should be exposed (public) from the `pkg/docker_exec_ws` package, right?

But I agree, having them public and documented is nice... so a subpackage might be a good way to hide them from people using the package to reduce confusion...

external/shellclient/v1/infinitepipe.go
```diff
@@ -0,0 +1,112 @@
...
+
+// InfinitePipe is similar to ioext.AsyncPipe() except that writes will always
+// succeed. Data will be added to an internal buffer that can grow bigger than
+// the specified capacity.
+// Additionally, you may supply a channel tell that will be told whenever
+// the Infinite channel has been emptied, so that more bytes can be requested
+// to be written.
+//
+// This pipe kind is useful when implementing simple congestion control.
+func InfinitePipe(tell chan<- bool) (*InfinitePipeReader, *InfinitePipeWriter) {
```
jonasfj: So this pipe never does `tell<- ...` anywhere....

So I suspect you have two pipes because:

A) For stdout/stderr (ie. the incoming streams), we need a to write the messages to a pipe that has an infinite internal buffer, and when the buffer is above a certain threshold you want to send "pause", when the buffer is empty you want to send "resume" with the respective `streamID == stdout or stderr`.

It seems to me that `DrainingPipe` is good for this, right? When you receive a message from `tell` the pipe is empty and we should send a "resume" message... If when writing to `DrainingPipe` you get a `ErrPipeFull` error then you should send a "pause" message.

B) For stdin (ie. the outgoing stream), we probably need a pipe we can block... Say we call it `r, w := BlockablePipe()`.
Then it should always be safe to call `r.Block()` to block further reads from the pipe, and `r.Unblock()` to unblock further reads...
In unblocked mode the pipe just acts like a normal pipe... In blocked mode writes are blocked.

So you would have one thread reading from `r` and writing messages to the websocket... and if in the receiving thread you get a message to "pause" stdin, then you call `r.Block()` if you get a message to "resume" stdin, then you call `r.Unblock()`.

I think you could do this by simply wrapping an existing `io.Pipe()` and then have a `sync.Cond/sync.Mutex` and a `bool` indicating if it's blocked or not...

external/shellclient/v1/shellclient.go
```diff
@@ -0,0 +1,352 @@
...
+	outputChan := make(chan bool)
+	stdout, stdoutWriter := DrainingPipe(MaxOutstandingBytes, outputChan)
+	stderr, stderrWriter := DrainingPipe(MaxOutstandingBytes, outputChan)
+
+	s := &ShellClient{
+		ws:      ws,
+		done:    make(chan struct{}),
+		success: true,
+
+		cPause: sync.NewCond(&sync.Mutex{}),
```
jonasfj: Please create it as `m sync.Mutex` and `c sync.Cond` like you did in the pipes it's a lot prettier... and probably you can use just one mutex/cond and the entire object... that keeps it simpler even if they protect two variables (so long as it doesn't create a deadlock it's a good idea to have as few locks as possible).

external/shellclient/v1/drainingpipe.go
```diff
@@ -0,0 +1,134 @@
+package v1
```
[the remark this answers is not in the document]
t0xicCode: done!

external/shellclient/v1/drainingpipe.go
```diff
@@ -0,0 +1,134 @@
...
+	m        sync.Mutex
+	buffer   []byte
+	draining bool
+	closed   bool
+	broken   bool
+	tell     chan<- bool
+	capacity int
+}
+
+// DrainingPipe is similar to ioext.AsyncPipe() except that writes will always
```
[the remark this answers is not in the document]
t0xicCode: done!

cmds/shell/cmd.go
```diff
@@ -83,7 +84,11 @@ func Execute(cmd *cobra.Command, args []string) error {
...
 	case "1":
-		return errors.New("the shell client doens't yet support v1 shells")
+		sockURL, _ = url.Parse(redirectURL.Query().Get("socketUrl"))
+		shell, err = v1client.Dial(sockURL.String(), []string{"bash"}, tty)
+		if err != nil {
+			return fmt.Errorf("could not create the shell client: %v", err)
+		}
 	case "2":
 		sockURL, _ = url.Parse(redirectURL.Query().Get("socketUrl"))
 		shell, err = v2client.Dial(sockURL.String(), []string{"bash"}, tty)
```
jonasfj: The v2 protocol is defined to start the default system shell, if no command is given...
So if instead of giving `[]string{"bash"}` we pass `nil` or []string{}` (I think either would work), we should get the default system shell that the machine is configured to provide... hence, `bash` or `sh` on linux and `cmd.exe` on windows...

Also we should probably tweak this command to take command arguments, instead of:
```
taskcluster shell <taskId>
```
We do something like:
```
taskcluster shell <taskId> [--] [command...]
```
So that all these examples would work:
```sh
# Various ways to get an interactive, the first one always works
# because it gets system shell... the other work depending on what is installed
# on the remote system.
taskcluster shell $TASKID                           # Get default system shell
taskcluster shell $TASKID -- bash                   # Force starting bash from PATH
taskcluster shell $TASKID -- bash --login           # provide argument
taskcluster shell $TASKID -- sh                     # Try sh from PATH (on remote system)
taskcluster shell $TASKID -- python                 # Or what about python
taskcluster shell $TASKID -- /usr/bin/env python    # we could even do it the right way
# We could also use this to run commands, to just deplay some data:
taskcluster shell $TASKID -- lsb_release --all      # Just run a command on the remote system
taskcluster shell $TASKID -- cat /var/log/syslog    # Dump system log
taskcluster shell $TASKID -- ps aux                 # Show programs running
# Or we could dump system log as it is written
taskcluster shell $TASKID -- tail -f /var/log/syslog
```


This way the `taskcluster shell` command allows for a lot more than just getting a simple shell... at the same time it's super easy to get an interactive shell :)

Note: The `--` in `taskcluster shell $TASKID -- ` means the arguments following `--` are all positional.. hence, when we do: `taskcluster shell $TASKID -- lsb_release --all` the `--all` flag won't be interpreted as a flag by `taskcluster shell` it'll be read as an argument that is part of the command array... I'm not sure how cobra does this, but `--` is a GNU convention, so it probably supports this :)

jonasfj: Since v1 protocol doesn't provide the default system shell, if no command is given... I think it just doesn't work if no command is given...

I propose that if we are in protocol v1 (legacy) and no command is given we use the default command that the `shell.html` tool on taskcluster-tools uses:
https://github.com/taskcluster/taskcluster-tools/blob/edb67d523a8302313b1046448d7cbfda33c6d196/src/views/Shell/Shell.js#L7-L24

Yes, that's a single command with a script as the last argument :) hehe
We can just embed as a constant for use in protocol v1 case (when no command is given).

cmds/shell/cmd.go
```diff
@@ -83,7 +84,11 @@ func Execute(cmd *cobra.Command, args []string) error {
 
 	switch redirectURL.Query().Get("v") {
 	case "1":
-		return errors.New("the shell client doens't yet support v1 shells")
+		sockURL, _ = url.Parse(redirectURL.Query().Get("socketUrl"))
+		shell, err = v1client.Dial(sockURL.String(), []string{"bash"}, tty)
```
jonasfj: See comment longer down wrt. to command given...

external/pkg/docker-exec-ws/doc.go
```diff
@@ -0,0 +1,4 @@
+// Package dockerExecWS provides a golang client implementation for the websocket protocol implemented in
+// the npm package "docker-exec-websocket-server". This is a websocket protocol for exposing "docker exec" to
+// a remote process.
+package dockerExecWS
```
jonasfj: how about `package dockerexecws`, oh, that's not much better... Can we come up with a better package name or should we just give up?

See: https://blog.golang.org/package-names

cmds/shell/cmd.go
```diff
@@ -15,6 +15,7 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/taskcluster/taskcluster-cli/cmds/root"
 	"github.com/taskcluster/taskcluster-cli/config"
+	v1client "github.com/taskcluster/taskcluster-cli/external/pkg/docker-exec-ws"
```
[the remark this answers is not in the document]
jonasfj: note, if there is convention arguing for why should put it in `external/` I'm totally open to doing that...

external/pkg/docker-exec-ws/doc.go
jonasfj: I'm mostly thinking we want to avoid problems with metalinter, I filed #171 to update the config and run more linters, but that shouldn't block this work :)

external/pkg/docker-exec-ws/shellclient.go
```diff
@@ -0,0 +1,353 @@
...
+
+	// Ensure the URL has ws or wss as scheme
+	switch u.Scheme {
+	case "http":
+		u.Scheme = "ws"
+	case "https":
+		u.Scheme = "wss"
+	} // Set command arguments overwriting any existing querystring values
+	q.Del("command")
+	if len(command) > 0 {
```
jonasfj: you don't need this "if", do you... the loop will just be a no-op if there is zero...

external/pkg/docker-exec-ws/shellclient.go
```diff
@@ -0,0 +1,353 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Return a new ShellClient wrapping the websocket
+	return New(ws), nil
+}
+
+// New creates a new v1 ShellClient from an existing websocket.
+func New(ws *websocket.Conn) *ShellClient {
```
jonasfj: Should this be public?

external/pkg/docker-exec-ws/shellclient.go
```diff
@@ -0,0 +1,353 @@
...
+	stdin        io.WriteCloser
+	stdinReader  io.ReadCloser
+	stdout       io.ReadCloser
+	stdoutWriter *iopipes.DrainingPipeWriter
+	stderr       io.ReadCloser
+	stderrWriter *iopipes.DrainingPipeWriter
+}
+
+// Dial will built a proper shell websocket URL, connect to the server and return an initialized shell client.
+func Dial(socketURL string, command []string, tty bool) (*ShellClient, error) {
```
jonasfj: Maybe I suggest that we add an additional optional argument `logger`... like:

```go
// Logger defines the interface for writing log messages, implemented by logrus and others..
type Logger interface {
  Debugf(format string, a ...interface{})
  Infof(format string, a ...interface{})
  Warnf(format string, a ...interface{})
  Errorf(format string, a ...interface{})
}

// A trivial logger implementation that discards all messages
type nilLogger struct{}
func (nilLogger) Debugf(string, ...interface{}) {}
func (nilLogger) Infof(string, ...interface{}) {}
func (nilLogger) Warnf(string, ...interface{}) {}
func (nilLogger) Errorf(string, ...interface{}) {}

//.... logger is optional
func Dial(socketURL string, command []string, tty bool, logger Logger) (*ShellClient, error) {
  if logger == nil {
    logger = nilLogger{}
  }

}
```
Now instead of littering the code with `fmt.Fprintln(os.Stderr, "Resuming client")` which is nice for debugging, we can specify a logger which we use during tests and in development... and when we do production we just give `nil` as logger and messages go away...

We could even implement a verbose mode in the tool like: `taskcluster shell $TASKID --verbose [--] [command...]`, which would print some or all of these warning messages...

Note: the way I proposed the `Logger` interface it is already implemented by https://godoc.org/github.com/sirupsen/logrus
 and by https://github.com/taskcluster/taskcluster-worker/blob/79da11e34f8b2e7ef8a01c101749717e00b1666b/runtime/monitor.go
which is used in tc-worker... This is kind of a neat feature with golang...

cmds/shell/cmd.go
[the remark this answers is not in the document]
t0xicCode: I see. I had put it in external because I intended to eventually move it to a separate repository, but it can live in `pkg/` too. Moved!

cmds/shell/cmd.go
```diff
@@ -83,7 +84,11 @@ func Execute(cmd *cobra.Command, args []string) error {
 
 	switch redirectURL.Query().Get("v") {
 	case "1":
-		return errors.New("the shell client doens't yet support v1 shells")
+		sockURL, _ = url.Parse(redirectURL.Query().Get("socketUrl"))
+		shell, err = v1client.Dial(sockURL.String(), []string{"bash"}, tty)
+		if err != nil {
+			return fmt.Errorf("could not create the shell client: %v", err)
```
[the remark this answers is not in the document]
t0xicCode: Looks interesting! Is it something we'd like to implement across the entire codebase? If so, would it make sense to implement it in a separate PR?

t0xicCode:
@jonasfj I have a few more of the comments to go through and I want to add some testing too.

external/pkg/docker-exec-ws/shellclient.go
t0xicCode: done

cmds/shell/cmd.go
```diff
@@ -81,12 +83,42 @@ func Execute(cmd *cobra.Command, args []string) error {
...
+
 	switch redirectURL.Query().Get("v") {
 	case "1":
-		return errors.New("the shell client doens't yet support v1 shells")
+		if len(command) == 0 {
+			// Default command for v1
+			// as defined in https://github.com/taskcluster/taskcluster-tools/blob/edb67d523a8302313b1046448d7cbfda33c6d196/src/views/Shell/Shell.js#L7-L24
+			command = []string{
+				"sh", "-c",
+				strings.Join([]string{
```
jonasfj: You can make multi-line strings with in golang:
```go
const myLongString = `
blablabla
new line
something more
`

```

pkg/docker-exec-ws/shellclient.go
```diff
@@ -0,0 +1,355 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Return a new ShellClient wrapping the websocket
+	return New(ws), nil
+}
+
+// New creates a new v1 ShellClient from an existing websocket.
+func New(ws *websocket.Conn) *ShellClient {
```
jonasfj: Should we expose this? when `Dial` is the interface people should use?

cmds/shell/cmd.go
t0xicCode: True, I made it as a `strings.Join()` so that it would generate a single line but still be readable in source code :)

pkg/docker-exec-ws/shellclient.go
t0xicCode: I left it exposed because it is to match the v2 client (https://godoc.org/github.com/taskcluster/taskcluster-worker/plugins/interactive/shellclient) that has it exposed.

jonasfj: oh, yeah, I guess it might be useful from time to time...

I think I needed in the worker, because I do some really crazy things with websockets :)

Issue #156, opened 2017-04-13
APIs accepting continuation tokens don't work


djmitche:
```
$ ./taskcluster api pulse listNamespaces
Error: Request failed: Non-2xx StatusCode: 400 received in 1 attempts
```
from the logs:
```
2017-04-13T21:56:12.570496+00:00 heroku[router]: at=info method=GET path="/v1/namespaces?continuation=&limit=" host=pulse.taskcluster.net request_id=3162b327-df26-4325-b6c1-848a680707d4 fwd="54.244.107.133" dyno=web.1 connect=1ms service=32ms status=400 bytes=860 protocol=https
```
and indeed:
```
$ curl https://pulse.taskcluster.net/v1/namespaces?limit=
{
  "code": "InvalidRequestArguments",
  "message": "Query-string parameter: limit=\"\" does not match expression: /[0-9]+/\n----\nmethod:     listNamespaces\nerrorCode:  InvalidRequestArguments\nstatusCode: 400\ntime:       2017-04-13T21:56:58.600Z",
  "requestInfo": {
    "method": "listNamespaces",
    "params": {},
    "payload": {},
    "time": "2017-04-13T21:56:58.600Z"
  }
}
```

t0xicCode:
Will add some testing and close this.

PR #172, opened 2017-10-26
Only add query parameter if non-empty

srfraser:
Addresses #156 allowing the use of listLatestArtifacts and similar.


t0xicCode:
Awesome!

PR #174, opened 2017-11-06
Bug 1395349 - refactor signin to use new client creator, don't store

djmitche:
It's best to encourage users to not keep TC credentials on disk, so this
writes them out as env vars instead.  The result will still *read* from
`~/.config/taskcluster.yml`, just not write to it.

Example:

```
dustin@jemison $ eval $(./taskcluster signin --scope 'queue:*')
Starting
Listening for a callback on: http://localhost:33459
Opening URL: https://tools.taskcluster.net/auth/clients/new?name=cli&description=Temporary+client+for+use+on+the+command+line&scope=queue:*&expires=1d&callback_url=http%3A%2F%2Flocalhost%3A33459
Credentials recorded as environment variables

dustin@jemison $ ./taskcluster api auth currentScopes
{
  "scopes": [
    "queue:*"
  ]
}
```

### DIFF
--- a/cmds/signin/signin.go
+++ b/cmds/signin/signin.go
@@ -20,24 +20,36 @@ import (
 func init() {
 	cmd := &cobra.Command{
 		Use:   "signin",
-		Short: "Obtain temporary credentials from login.taskcluster.net.",
-		Long: `The command 'taskcluster signin' will open your web-browser to
-login.taskcluster.net where you can sign-in and obtain temporary
-credentials.
+		Short: "Get Taskcluster credentials and export to the shell",
+		Long: `The command 'taskcluster signin', run on a desktop system, will use
+your browser to get Taskcluster credentials for use with other commands.
 
-Once signed in, you can click the 'Grant Access' button which
-will redirect you to localhost where this command will be listening
-and save the temporary credentials to local configuration file.`,
+Use it like this:
+
+$ eval ` + "`taskcluster signin`" + `
+
+You might make this easy to use with an alias in ~/.bashrc:
+
+alias tc-signin="eval ` + "`taskcluster signin`" + `"
+
+This will set environment variables in your shell session containing the credentials.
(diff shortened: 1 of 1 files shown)


cmds/signin/signin.go
```diff
@@ -111,17 +109,23 @@ func cmdSignin(cmd *cobra.Command, _ []string) error {
...
+		loginURL += "&scope=" + scopes[i]
+	}
+	expires, _ := cmd.Flags().GetString("expires")
+	loginURL += "&expires=" + url.QueryEscape(expires)
+	loginURL += "&callback_url=" + url.QueryEscape(callbackURL)
 
 	// Open browser
-	fmt.Fprintln(cmd.OutOrStdout(), "Listening for a callback on: "+target)
-	fmt.Fprintln(cmd.OutOrStdout(), "Opening URL: "+loginURL)
+	fmt.Fprintln(cmd.OutOrStderr(), "Listening for a callback on: "+callbackURL)
```
jonasfj: So one would write: 
```bash
alias taskcluster-login="eval `taskcluster login`"
```

even better:
```bash
taskcluster() {
    if [[ $@ == "-la" ]]; then
        eval `taskcluster login`
    else
        taskcluster "$@"
    fi
}
```

jonasfj: Could we make an argument like `taskcluster login --setup` that will install this in `/etc/bash_completion.d/` or maybe we can integrate it with the bash completion we have.. (not sure if this abuse of bash_completion).

Or at-least have a section in `taskcluster help login` that explains how to create these and add them to `.bashrc`

djmitche: I don't want to mess with shell completion.  And that function needs some work to properly find the path to taskcluster and not invoke itself recursively (and also to use `taskcluster signin` not `taskcluster login`).  I think an alias is simpler.

PR #175, opened 2017-11-08
Fix Windows URL in README

t0xicCode:
Looks like the `.exe` is missing from the URL for the windows downloads in the `README` file.


djmitche:
derp.

Issue #181, opened 2017-11-24
Using taskcluster-cli cleanly within multiple shells


srfraser:
In the current incarnation, we need to copy/paste the two environment variables into each shell we'd like to use `taskcluster` in, since it doesn't store them on disk any more.

A clean way of using the cli in multiple shells would be very helpful. 

I know that the commit message says the client still reads from `taskcluster.yaml`, but since I deleted this file to try and work out some issues, I no longer have a reference copy to put those values in - I don't know the format.

How much of a benefit is there to avoid writing these temporary credentials to disk, given that they're temporary?

djmitche:
One option might be:

```
$ taskcluster signin > /dev/shm/creds && source /dev/shm/creds
```

and then in another shell, just run the `source` part.  You could even work that `source` into your `.bashrc` or equivalent, or maybe build some shell function that looks at the age of /dev/shm/creds and only runs `taskcluster signin` if they are too old, then sources them regardless.

If you come up with something nice, I'd be happy to include it in the docs.

It's OK if you decide to write to disk (noting that /dev/shm is *not* disk, but if that was `~/.tc-signin` that would be OK too).  I just don't want to do that by default.  There's risk of other processes reading the data (even my example above doesn't set file permissions correctly) and of an attacker capturing a powered-on laptop and extracting the secrets that way.  So it's more of a defense-in-depth thing.

djmitche:
@srfraser did #182 solve this?

PR #182, opened 2017-11-28
Add a scriptable check for valid credentials

srfraser:
This allows users to `taskcluster signin --check` and get
an exit code of 1 if their credentials are either not found,
or expired.

I've successfully used it with this bash wrapper:

    tc() {
        if ! TC=$(which taskcluster)
        then
            echo "Unable to find taskcluster command"
            return
        fi 
    
        CLIENTVARS=${TASKCLUSTER_CLIENT_STORE:-"/dev/shm/tcclient.creds"}
        if [ -f "$CLIENTVARS" ] ; then
            # shellcheck source=/dev/null
            source "$CLIENTVARS"
        fi

        if ! ${TC} signin --check; then
            ${TC} signin > "$CLIENTVARS"
            # shellcheck source=/dev/null
            source "$CLIENTVARS"
        fi
        ${TC} "$@"
    }

### DIFF
--- a/cmds/signin/signin.go
+++ b/cmds/signin/signin.go
@@ -7,6 +7,7 @@ import (
 	"net"
 	"net/http"
 	"net/url"
+	"os"
 	"strings"
 	"time"
 
@@ -14,6 +15,8 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/taskcluster/taskcluster-cli/cmds/root"
 	"github.com/taskcluster/taskcluster-cli/config"
+	tcclient "github.com/taskcluster/taskcluster-client-go"
+	"github.com/taskcluster/taskcluster-client-go/auth"
 	graceful "gopkg.in/tylerb/graceful.v1"
 )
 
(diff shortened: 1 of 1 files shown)


srfraser:
A potential option for https://github.com/taskcluster/taskcluster-cli/issues/181 (which I forgot to mention in the initial comment)

cmds/signin/signin.go
```diff
@@ -141,3 +149,23 @@ func cmdSignin(cmd *cobra.Command, _ []string) error {
...
+		creds = config.Credentials.ToClientCredentials()
+	}
+	auth := auth.New(creds)
+	result, err := auth.CurrentScopes()
+	if err != nil {
+		// Don't want an os.Exit() in case it causes scripting loops when used.
+		return fmt.Errorf("failed to check credentials: %s", err)
+	}
+	if len(result.Scopes) == 0 {
+		fmt.Println("No valid credentials were found")
```
djmitche: More accurate to say "No valid credentials were found, or credentials have no scopes".

I think this is a good way to check for access all the same, but if someone found themselves in the weird case of having a clientId with no scopes, the message might help them figure out what's going on.

djmitche:
BTW, in the shell wrapper it'd probably be good to have the creds in a subdirectory where that subdirectory is chmod 700.

PR #184, opened 2018-01-10
url escape scopes

jonasfj:
warning, I didn't test this yet..


jonasfj:
gps says it makes encoding look better... but it didn't solve the issue... probably a JS bug...

PR #192, opened 2018-06-04
Bug 1466679 - use rootUrl in 'taskcluster signin'

djmitche:


djmitche:
I'm not sure why the first commit changed anything in `vendor.json`.  Running `make` updated those hashes.  I can revert that file if it's the right thing to do.

djmitche:
The test failure is because the "1000 year task" apparently only had a 1-year artifact expiration.  I've started another (https://tools.taskcluster.net/groups/SmJVXbZERxigZVunstEveQ/tasks/SmJVXbZERxigZVunstEveQ/details) and will land a test fix if that works.

djmitche:
Rebased over the test fix

petemoore:
I'll have a look at this tomorrow!

djmitche:
@petemoore ping?

petemoore:
I'm looking at this now! :-)

djmitche:
Thanks!  We use the web client because it's not possible to sign in with Auth0 without a browser.

I don't see a reference to `graceful` here -- I didn't write the rest of the signin command, so maybe that was someone else.

PR #193, opened 2018-06-05
Bug 1194767 - Use "nice" slugs for taskIds

petemoore:
At the same time, I formatted the source using `go fmt`.

See [bug 1194767](https://bugzilla.mozilla.org/show_bug.cgi?id=1194767) for context.

Also see taskcluster/generic-worker#96.


petemoore:
Switching review to @djmitche, simply because I've just given him two other `taskcluster-cli` reviews, so it might be easier for him to tackle all three in one go rather than consuming time from both of you!

PR #194, opened 2018-06-05
Bug 1444458 - Expire cached ping URLs after 365 days rather than one day

petemoore:
@djmitche 
While we wait for a complete solution to [bug 1444458](https://bugzil.la/1444458), this is a partial solution so the ping urls cache only expires after 365 days, rather than after one day.


petemoore:
Thanks!

PR #195, opened 2018-06-05
Fix TestLogCommand

petemoore:
This just fixes the PR. I don't know of a way to configure the task log to expire at anything other than the default of one year in the future.

Maybe we should update the s3 bucket directly, so that this doesn't again expire in a year?

What do you think?

Incidentally, aren't there logs we would want to keep longer than a year? I hadn't really thought about this before - we probably have the same issue in generic-worker and taskcluster-worker too.


petemoore:
Thanks!

PR #196, opened 2018-06-06
Issue 188: Add noop flag for cancel/rerun/complete

mutterroland:
Added noop flag for cancel/rerun/complete commands.

### DIFF
--- a/cmds/task/actors.go
+++ b/cmds/task/actors.go
@@ -22,12 +22,12 @@ func runCancel(credentials *tcclient.Credentials, args []string, out io.Writer, 
 
 
 	if noop {
-		displayNoopMsg("Would cancel", credentials, args, out, flagSet)
+		displayNoopMsg("Would cancel", credentials, args)
 		return nil
 	}
 
 	if confirm {
-		var response = confirmMsg("Cancels", credentials, args, out, flagSet)
+		var response = confirmMsg("Cancels", credentials, args)
 		if response == "y" {
 			c, err := q.CancelTask(taskID)
 			run := c.Status.Runs[len(c.Status.Runs)-1]
@@ -66,12 +66,12 @@ func runRerun(credentials *tcclient.Credentials, args []string, out io.Writer, f
 	taskID := args[0]
 
 	if noop {
-		displayNoopMsg("Would re-run", credentials, args, out, flagSet)
+		displayNoopMsg("Would re-run", credentials, args)
 		return nil
 	}
 
 	if confirm {
-		var response = confirmMsg("Will re-run", credentials, args, out, flagSet)
+		var response = confirmMsg("Will re-run", credentials, args)
 		if response == "y" {
 			c, err := q.RerunTask(taskID)
 			run := c.Status.Runs[len(c.Status.Runs)-1]
@@ -212,7 +212,7 @@ func runComplete(credentials *tcclient.Credentials, args []string, out io.Writer
 
 
 	if confirm {
-		var response = confirmMsg("Will complete", credentials, args, out, flagSet)
+		var response = confirmMsg("Will complete", credentials, args)
 		if response == "y" {
 			c, err := q.ClaimTask(taskID, fmt.Sprint(len(s.Status.Runs)-1), &queue.TaskClaimRequest{
 				WorkerGroup: s.Status.WorkerType,
@@ -238,7 +238,7 @@ func runComplete(credentials *tcclient.Credentials, args []string, out io.Writer
 	}
 
 	if noop {
-		displayNoopMsg("Would complete", credentials, args, out, flagSet)
+		displayNoopMsg("Would complete", credentials, args)
 		return nil
 	}
 
--- a/cmds/task/actors.go
+++ b/cmds/task/actors.go
@@ -5,7 +5,7 @@ import (
 	"io"
 	"os"
 	"time"
-	
+
 	"github.com/spf13/pflag"
 	"github.com/taskcluster/slugid-go/slugid"
 	tcclient "github.com/taskcluster/taskcluster-client-go"
@@ -14,19 +14,19 @@ import (
 
 // runCancel cancels the runs of a given task.
 func runCancel(credentials *tcclient.Credentials, args []string, out io.Writer, flagSet *pflag.FlagSet) error {
-	nooprunCancel, _ := flagSet.GetBool("noop")
-	confirmrunCancel, _ := flagSet.GetBool("confirm")
+	noop, _ := flagSet.GetBool("noop")
+	confirm, _ := flagSet.GetBool("confirm")
 
 	q := makeQueue(credentials)
 	taskID := args[0]
 
 
-	if nooprunCancel {
-		displayNoopMsg("Cancels", credentials, args, out, flagSet)
+	if noop {
+		displayNoopMsg("Would cancel", credentials, args, out, flagSet)
 		return nil
 	}
 
-	if confirmrunCancel {
+	if confirm {
 		var response = confirmMsg("Cancels", credentials, args, out, flagSet)
 		if response == "y" {
 			c, err := q.CancelTask(taskID)
@@ -43,7 +43,7 @@ func runCancel(credentials *tcclient.Credentials, args []string, out io.Writer, 
 
 	c, err := q.CancelTask(taskID)
 	run := c.Status.Runs[len(c.Status.Runs)-1]
-		
+
 
 	if err != nil {
 		fmt.Fprintln(os.Stderr, err)
@@ -59,19 +59,19 @@ func runCancel(credentials *tcclient.Credentials, args []string, out io.Writer, 
 
 // runRerun re-runs a given task.
 func runRerun(credentials *tcclient.Credentials, args []string, out io.Writer, flagSet *pflag.FlagSet) error {
-	nooprunRerun, _ := flagSet.GetBool("noop")
-	confirmrunRerun, _ := flagSet.GetBool("confirm")
+	noop, _ := flagSet.GetBool("noop")
+	confirm, _ := flagSet.GetBool("confirm")
 
 	q := makeQueue(credentials)
 	taskID := args[0]
 
-	if nooprunRerun {
-		displayNoopMsg("Re-runs", credentials, args, out, flagSet)
+	if noop {
+		displayNoopMsg("Would re-run", credentials, args, out, flagSet)
 		return nil
 	}
-	
-	if confirmrunRerun {
-		var response = confirmMsg("Re-runs", credentials, args, out, flagSet)
+
+	if confirm {
+		var response = confirmMsg("Will re-run", credentials, args, out, flagSet)
 		if response == "y" {
 			c, err := q.RerunTask(taskID)
 			run := c.Status.Runs[len(c.Status.Runs)-1]
@@ -88,14 +88,14 @@ func runRerun(credentials *tcclient.Credentials, args []string, out io.Writer, f
 	c, err := q.RerunTask(taskID)
 
 	run := c.Status.Runs[len(c.Status.Runs)-1]
-	
+
 
 	if err != nil {
 		return fmt.Errorf("could not rerun the task %s: %v", taskID, err)
 
 	}
 
-	
+
 	fmt.Fprintln(out, getRunStatusString(run.State, run.ReasonResolved))
 	return nil
 
@@ -121,7 +121,7 @@ func runRetrigger(credentials *tcclient.Credentials, args []string, out io.Write
 
 	exactRetrigger, _ := flagSet.GetBool("exact")
 
-	newTaskID := slugid.V4()
+	newTaskID := slugid.Nice()
 	now := time.Now().UTC()
 
 	origCreated, err := time.Parse(time.RFC3339, t.Created.String())
@@ -181,8 +181,8 @@ func runRetrigger(credentials *tcclient.Credentials, args []string, out io.Write
 
 // runComplete completes a given task.
 func runComplete(credentials *tcclient.Credentials, args []string, out io.Writer, flagSet *pflag.FlagSet) error {
-	nooprunComplete, _ := flagSet.GetBool("noop")
-	confirmrunComplete, _ := flagSet.GetBool("confirm")
+	noop, _ := flagSet.GetBool("noop")
+	confirm, _ := flagSet.GetBool("confirm")
 
 	q := makeQueue(credentials)
 	taskID := args[0]
@@ -210,9 +210,9 @@ func runComplete(credentials *tcclient.Credentials, args []string, out io.Writer
 		return fmt.Errorf("could not complete the task %s: %v", taskID, err)
 	}
 
-	
-	if confirmrunComplete {
-		var response = confirmMsg("Completes", credentials, args, out, flagSet)
+
+	if confirm {
+		var response = confirmMsg("Will complete", credentials, args, out, flagSet)
 		if response == "y" {
 			c, err := q.ClaimTask(taskID, fmt.Sprint(len(s.Status.Runs)-1), &queue.TaskClaimRequest{
 				WorkerGroup: s.Status.WorkerType,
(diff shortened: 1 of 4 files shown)


cmds/task/actors.go
```diff
@@ -5,42 +5,74 @@ import (
...
+func runRerun(credentials *tcclient.Credentials, args []string, out io.Writer, flagSet *pflag.FlagSet) error {
+	noopRerun, _ := flagSet.GetBool("noop")
+
 	q := makeQueue(credentials)
 	taskID := args[0]
 
+	if noopRerun {
+		c, _ := q.Status(taskID)
+		run := c.Status.Runs[len(c.Status.Runs)-1]
+		fmt.Println("Rerunns", taskID ,"(state:",run.State,")")
```
srfraser: Spelling nit: 'Reruns'

cmds/task/actors.go
```diff
@@ -5,42 +5,74 @@ import (
...
+
 	q := makeQueue(credentials)
 	taskID := args[0]
 
+
+	if nooprunCancel {
+		c, _ := q.Status(taskID)
+		run := c.Status.Runs[len(c.Status.Runs)-1]
+		fmt.Println("Cancels", taskID ,"(state:",run.State,")")
+		runName(credentials , args , out , flagSet)
```
srfraser: It's not clear why this is a separate function call. Should we have the task name in the `Println` above? Having all the information on one line should help us if we run the `taskcluster` command inside scripts, or want to `grep` the output. 

There are options here:
1. We could take the contents of `runName` and move them into this function instead, giving us the name.
2. We could make a different version of `runName` that uses `return` to give us the name, which we can then use.
3. We could make a new function that prints out a `noop` message, and uses one of option 1 or 2. This lets us re-use the same code over and over for the no-op messages, and only change the format of them in one place.

For option 3, imagine we have:

    if nooprunCancel {
        displayNoopMsg(taskID, "Cancels", out)
    }

`func displayNoopMsg` can then look up all the information and have `fmt.Println(actionMsg, taskID, ...` 

What do you think?

mutterroland: Sounds awesome. Will try option 1. and 2. as soon as possible.

srfraser:
Looks good! Let's get a review from one of the maintainers

cmds/task/task.go
```diff
@@ -38,6 +55,14 @@ func init() {
 
 	retriggerCmd.Flags().BoolP("exact", "e", false, "Retrigger in exact mode. WARNING: THIS MAY HAVE SIDE EFFECTS. USE AFTER YOU READ THE SOURCE CODE.")
 
+	rerunCmd.Flags().BoolP("noop","n", false, "Specifies the operation to perform and adjust some operations to print extra information.")
```
srfraser: Language nit: `--noop` does not specify the operation to perform. It tells the command to not actually run the command, but print out what it would do

cmds/task/task.go
```diff
@@ -38,6 +55,14 @@ func init() {
...
+	rerunCmd.Flags().BoolP("confirm","c", false, "Prompts user with a confirmation (y/n) before performing any changes and prints extra informations.")
```
srfraser: Nit, I'd remove " and prints extra informations."

cmds/task/fetchers.go
```diff
@@ -58,7 +58,45 @@ func runStatus(credentials *tcclient.Credentials, args []string, out io.Writer,
 	fmt.Fprintln(out, getRunStatusString(s.Status.Runs[runID].State, s.Status.Runs[runID].ReasonResolved))
 	return nil
 }
+// confirmMsg displays confirmation message when --confirm is used
+func confirmMsg(command string, credentials *tcclient.Credentials, args []string, out io.Writer, _ *pflag.FlagSet )(string) {
```
srfraser: Nice code re-use!

cmds/task/actors.go
```diff
@@ -149,6 +210,38 @@ func runComplete(credentials *tcclient.Credentials, args []string, out io.Writer
 		return fmt.Errorf("could not complete the task %s: %v", taskID, err)
 	}
 
+	
+	if confirmrunComplete {
+		var response = confirmMsg("Completes", credentials, args, out, flagSet)
+		if response == "y" {
```
srfraser: If this is the same in each call to `confirmMsg`, the check for `y`, `Y`, `Yes` and other variants could be in the `confirmMsg` function

mutterroland: The variable is a different one in each case.

lundjordan:
@djmitche ^ I believe you are a maintainer for this repo and the appropriate person to review. sfraser gave it a pass and now @mutterroland's patch is ready to land. Could you review it when time permits?

djmitche:
Sure thing.  Thanks for doing the initial reviews @srfraser

cmds/task/actors.go
```diff
@@ -5,42 +5,100 @@ import (
...
 	"github.com/taskcluster/slugid-go/slugid"
 	tcclient "github.com/taskcluster/taskcluster-client-go"
 	"github.com/taskcluster/taskcluster-client-go/queue"
 )
 
 // runCancel cancels the runs of a given task.
-func runCancel(credentials *tcclient.Credentials, args []string, out io.Writer, _ *pflag.FlagSet) error {
+func runCancel(credentials *tcclient.Credentials, args []string, out io.Writer, flagSet *pflag.FlagSet) error {
+	nooprunCancel, _ := flagSet.GetBool("noop")
+	confirmrunCancel, _ := flagSet.GetBool("confirm")
```
djmitche: capitalization is confusing here - how about just `noop` and `confirm`?

cmds/task/fetchers.go
```diff
@@ -58,7 +58,45 @@ func runStatus(credentials *tcclient.Credentials, args []string, out io.Writer,
...
+	fmt.Println(command, t.Metadata.Name ,"taskid:", taskID ,"(state:",run.State,"). Are you sure you want to proceed with given command? (y/N)")
+	var response string
+	fmt.Scanf("%s", &response)
+
+
+	return response
+}
+
+// displayNoopMsg displays details when --noop is used
+func displayNoopMsg(command string, credentials *tcclient.Credentials, args []string, out io.Writer, _ *pflag.FlagSet ) error {
```
djmitche: like `confirmMsg`, there's no need to pass unused parameters here (`out` and the FlagSet), and it would be better to pass the taskId directly, rather than args.

cmds/task/fetchers.go
```diff
@@ -58,7 +58,45 @@ func runStatus(credentials *tcclient.Credentials, args []string, out io.Writer,
...
+	c, _ := q.Status(taskID)
+	run := c.Status.Runs[len(c.Status.Runs)-1]
+
+
+	t, err := q.Task(taskID)
+	if err != nil {
+		return fmt.Errorf("could not get the task %s: %v", taskID, err)
+	}
+
+	fmt.Println(command, t.Metadata.Name ,"taskid:", taskID ,"(state:",run.State,")")
```
djmitche: some whitespace would probably be good here? Perhaps fmt.Printf?

cmds/task/fetchers.go
```diff
@@ -58,7 +58,45 @@ func runStatus(credentials *tcclient.Credentials, args []string, out io.Writer,
...
+	run := c.Status.Runs[len(c.Status.Runs)-1]
+
+	t, _ := q.Task(taskID)
+	
+	fmt.Println(command, t.Metadata.Name ,"taskid:", taskID ,"(state:",run.State,"). Are you sure you want to proceed with given command? (y/N)")
+	var response string
+	fmt.Scanf("%s", &response)
+
+
+	return response
```
djmitche: It would be nice to have this function do the response interpretation and return a boolean (and also allow y or Y).

cmds/task/fetchers.go
djmitche: No need for parentheses around `string` here

cmds/task/fetchers.go
```diff
@@ -58,7 +58,45 @@ func runStatus(credentials *tcclient.Credentials, args []string, out io.Writer,
...
+	
+	q := makeQueue(credentials)
+	taskID := args[0]
+
+	c, _ := q.Status(taskID)
+	run := c.Status.Runs[len(c.Status.Runs)-1]
+
+	t, _ := q.Task(taskID)
+	
+	fmt.Println(command, t.Metadata.Name ,"taskid:", taskID ,"(state:",run.State,"). Are you sure you want to proceed with given command? (y/N)")
```
djmitche: `fmt.Printf` would be better here, as you can omit the final `\n` so that the cursor appears on the same line as the question.

cmds/task/actors.go
djmitche: it's always "y'..

cmds/task/actors.go
```diff
@@ -63,7 +121,7 @@ func runRetrigger(credentials *tcclient.Credentials, args []string, out io.Write
 
 	exactRetrigger, _ := flagSet.GetBool("exact")
 
-	newTaskID := slugid.Nice()
+	newTaskID := slugid.V4()
```
djmitche: Please don't revert this

cmds/task/fetchers.go
djmitche: Also, `command` here seems to be a capitalized continuing-present-tense verb like "Completes".  Since this is a no-op, perhaps something like `"Would complete %s - taskId %s, state %s"` would be a bit better.  This variable might be better named `action` or `verb` in this case.

cmds/task/actors.go
```diff
@@ -122,7 +180,10 @@ func runRetrigger(credentials *tcclient.Credentials, args []string, out io.Write
 }
 
 // runComplete completes a given task.
-func runComplete(credentials *tcclient.Credentials, args []string, out io.Writer, _ *pflag.FlagSet) error {
+func runComplete(credentials *tcclient.Credentials, args []string, out io.Writer, flagSet *pflag.FlagSet) error {
+	nooprunComplete, _ := flagSet.GetBool("noop")
+	confirmrunComplete, _ := flagSet.GetBool("confirm")
```
djmitche: same here - `noop` and `confirm`

cmds/task/actors.go
```diff
@@ -5,42 +5,100 @@ import (
...
+
+
 	fmt.Fprintln(out, getRunStatusString(run.State, run.ReasonResolved))
 	return nil
 }
 
+
 // runRerun re-runs a given task.
-func runRerun(credentials *tcclient.Credentials, args []string, out io.Writer, _ *pflag.FlagSet) error {
+func runRerun(credentials *tcclient.Credentials, args []string, out io.Writer, flagSet *pflag.FlagSet) error {
```
djmitche: similar comments for `runCancel` apply here and to `runComplete`.

cmds/task/run.go
```diff
@@ -84,7 +84,7 @@ func runRunTask(cmd *cobra.Command, args []string) error {
 	}
 
 	// Generate a new taskID
-	taskID := slugid.Nice()
+	taskID := slugid.V4()
```
djmitche: As above, please leave this using `Nice`.

cmds/task/actors.go
```diff
@@ -5,42 +5,100 @@ import (
...
+	confirmrunCancel, _ := flagSet.GetBool("confirm")
+
 	q := makeQueue(credentials)
 	taskID := args[0]
 
+
+	if nooprunCancel {
+		displayNoopMsg("Cancels", credentials, args, out, flagSet)
+		return nil
+	}
```
djmitche: This could occur before calling `makeQueue`.

cmds/task/fetchers.go
djmitche: No need to include unused parameters here - `out` and the FlagSet are unused.  I'm not sure why the linter didn't catch that..

Also, it would be preferable to pass the taskId directly, rather than passing args.  Then this function doesn't have to know about the structure of the args.  Everywhere this function is called already has the taskId.

cmds/task/run.go
mutterroland: I don't know how that happened. Haven't touch that ...

cmds/task/fetchers.go
mutterroland: Won't work like that :(

djmitche: Sorry, why not?

djmitche: I'm a little dubious that adding whitespace to a message causes "crazy"..

cmds/task/actors.go
```diff
@@ -13,34 +13,92 @@ import (
...
+		displayNoopMsg("Would cancel", credentials, args)
+		return nil
+	}
+
+	if confirm {
+		var response = confirmMsg("Cancels", credentials, args)
+		if response == "y" {
+			c, err := q.CancelTask(taskID)
+			run := c.Status.Runs[len(c.Status.Runs)-1]
+			if err != nil {
```
djmitche: this err check should come before trying to use `c` (which is probably nil in the error case, meaning the previous line will panic)

cmds/task/actors.go
```diff
@@ -13,34 +13,92 @@ import (
...
+			return nil
+		}
+		return nil
+	}
+
 	c, err := q.CancelTask(taskID)
+	run := c.Status.Runs[len(c.Status.Runs)-1]
+
+
 	if err != nil {
```
djmitche: Hm, this code looks exactly like the code above -- that could be modified to only write this code once (by returning early if confirmation is declined)

cmds/task/actors.go
```diff
@@ -149,6 +210,38 @@ func runComplete(credentials *tcclient.Credentials, args []string, out io.Writer
...
+			if err != nil {
+				return fmt.Errorf("could not complete the task %s: %v", taskID, err)
+			}
+			fmt.Fprintln(out, getRunStatusString(r.Status.Runs[c.RunID].State, r.Status.Runs[c.RunID].ReasonResolved))
+			return nil
+
+		}
+	}
+
+	if noop {
```
djmitche: the noop check comes first for the other commands..

cmds/task/actors.go
```diff
@@ -149,6 +210,38 @@ func runComplete(credentials *tcclient.Credentials, args []string, out io.Writer
 		return fmt.Errorf("could not complete the task %s: %v", taskID, err)
 	}
 
+
```
djmitche: It looks like this performs the completion, *then* asks for confirmation, and if "y" runs the completion *agian*?  I don't think that's right!

cmds/task/fetchers.go
```diff
@@ -58,7 +58,45 @@ func runStatus(credentials *tcclient.Credentials, args []string, out io.Writer,
...
+
+	q := makeQueue(credentials)
+	taskID := args[0]
+
+	c, _ := q.Status(taskID)
+	run := c.Status.Runs[len(c.Status.Runs)-1]
+
+	t, _ := q.Task(taskID)
+
+	fmt.Println(command, t.Metadata.Name ,"taskid:", taskID ,"(state:",run.State,"). Are you sure you want to proceed with given command? (y/N)")
```
djmitche: This would be better with `fmt.Printf` and format specifiers.  If you leave the \n off the end, then the prompt will be on the last line, which is normal for confirmation prompts.  Something like `fmt.Printf("%s %s taskId: %s (state: %s). Are you sure you want to proceed? ", command, t.Metadata.Name, taskId, run.State)`.  Note the space after the `?`.

cmds/task/fetchers.go
```diff
@@ -58,7 +58,45 @@ func runStatus(credentials *tcclient.Credentials, args []string, out io.Writer,
...
+	c, _ := q.Status(taskID)
+	run := c.Status.Runs[len(c.Status.Runs)-1]
 
+
+	t, err := q.Task(taskID)
+	if err != nil {
+		return fmt.Errorf("could not get the task %s: %v", taskID, err)
+	}
+
+	fmt.Printf("%s %s taskid: %s (state: %s)\n", command, t.Metadata.Name, taskID, run.State)
```
djmitche: :heart:

cmds/task/actors.go
mutterroland: Maybe I'm not getting the idea but confirm is valid only if the user adds flag --confirm. If user doesn't add --confirm the task will cancel without user's confirmation.

djmitche: Right, the behavior is correct, but the same code (`q.CancelTask` and the lines around it) appears twice.  Duplicating code makes later maintenance difficult, because any change to that code must be made twice, and that's easily forgotten.

djmitche:
This is looking good :)

I pushed two more commits.  The first just removes some extra blank lines.  And my editor automatically applies `gofmt` so that made some other changes.

The second handles errors in confirmMsg that were previously ignored.  I should have spotted that on the last round of review, so I thought it was nicer to just fix it rather than send you another "changes requested".

It's often easier to ignore errors when hacking around, but I like to add something like `// TODO: handle error` so I remember to go add the error handling later, before requesting review :)

Anyway, with these minor changes in place, and the tests having passed, I'll get this merged.  Thanks!

PR #210, opened 2019-01-24
Update go-got package

walac:
This includes the fix to report body messages in case of request errors.


walac:
The error is unrelated to the patch

PR #212, opened 2019-07-01
Do not expire release binaries artifacts

SimonSapin:
Links under https://github.com/taskcluster/taskcluster-cli#installation are currently 404, presumably because it’s been more than a year (the default task expiry) since the last release.

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -137,6 +137,7 @@ tasks:
       taskId: {$eval: as_slugid("release_task")}
       created: {$fromNow: ''}
       deadline: {$fromNow: '1 hour'}
+      expires: {$fromNow: '100 years'}
       scopes:
         - auth:aws-s3:read-write:downloads-taskcluster-net/taskcluster-cli/
       routes:


djmitche:
Hm, they should be on downloads.taskcluster.net now.  I'll check..

SimonSapin:
That host name does not resolve in DNS.

djmitche:
Indeed - I'm obviously mis-remembering.  Maybe we switched the other direction?

djmitche:
Hm, this isn't great -- it doesn't build anymore because go.

How urgent is this?  Can I send you a copy of the linux-amd64 binary?  We'll soon be porting this to the monorepo at which point it will remain maintained :/

SimonSapin:
It’s not urgent. I already have a copy on my laptop, but I’m considering using this tool in a script that will go into https://github.com/servo/servo/. This script is only used manually when updating the Windows AMI so it’s not used often. And it’ll most likely be used by me, but I want other people on my team to be able to use it as well when I’m on vacation or something.

I’m considering this tool over the Python client (even though the script is written in Python) because `taskcluster signin` is convenient.

djmitche:
Yeah, that's part of why we're moving it in-tree.  I'll try to work on that soon.

